Issue #111, opened 2020-11-13
[3pt] Viz output requirements


LauraKeys-NOAA:
Viz team requesting post-processing of FIM 3 datasets as necessary to have the following outputs:

- hydrotable.csv rating curves reformatted as json files in following format: (values replacing <>)
{"<hydro id here>": {"stage_list": [0.0, 0.3048, 0.6096, <etc>], "q_list": [0.0, 9.311293553556311, 42.08217534559019, <etc>], "nwm_feature_id": <crosswalked nwm feature id here>}}
- catchments and hand output files renamed as previously discussed
- all HUC8 data in the above files compiled on a HUC6-basis

BradfordBates-NOAA:
Related to #94

BradfordBates-NOAA:
Related to #88

LauraKeys-NOAA:
- hydrotable.csv rating curves reformatted as json files in following format: (values replacing <>)
{"\<fim3 id_here\>": {"stage_list": [0.0, 0.3048, 0.6096, \<etc\>], "q_list": [0.0, 9.311293553556311, 42.08217534559019, \<etc\>], "nwm_feature_id": \<nwm feature_id here\>}}

- catchments and hand output files renamed with suggested format 
\<folder: fim3_library\>
 -> \<folder: mainstems\>
    --> \<folder: huc code\>
         ---> catchments_\<huc code\>.tif
         ---> rating_curves_\<huc code\>.json         
         ---> hand_grid_\<huc code\>.tif
 -> \<folder: full-resolution\>
     -->
         --->

- all HUC8 data in the above files compiled on a HUC6-basis

nickchadwick-noaa:
This may be a question for @BradfordBates-NOAA . I am not sure what files should be renamed to what in the outputs. From Brad, I got a list of "important files" in the outputs being:

- rem_zeroed_masked.tif
- gw_catchments_reaches_filtered_addedAttributes_crosswalked.gpkg
- gw_catchments_reaches_filtered_addedAttributes.tif
- hydroTable.csv
- src.json

Of those 5 files, which 3 are the ones that @LauraKeys-NOAA mention in her "suggested format".

Also, due to my lack of knowledge in hydrology, you're going to need to explain your last bullet point "all HUC8 data in the above files compiled on a HUC6-basis" a little bit more so I know how to go about formatting these files the way you want them.

The first bullet point of "hydrotable.csv rating curves reformatted as json..." should be straightforward.

[...]

Issue #94, opened 2020-11-03
[3pt] Simplify names of necessary-for-inundation files


BradfordBates-NOAA:
For example, change "gw_catchments_reaches_filtered_addedAttributes.tif" to "catchments.tif"

BradfordBates-NOAA:
@nickchadwick-noaa please get with @LauraKeys-NOAA and @CoreyKrewson-NOAA to decide on these file names.

nickchadwick-noaa:
This issue is being addressed in #88

PR #145, opened 2020-12-09
[1pt] Added cleanup module for production runs and viz post-processing

### LINKED ISSUE #88: [8pt] Create "janitor" script in Python to delete all unnecessary-for-inundation files from the fim_run.sh output directory.
There are several intermediate files produced by fim_run.sh. The majority of these files are not necessary for inundation mapping. This takes up storage space. The janitor script will clear out these files. The script would be called at the very end of run_by_unit.sh and a flag will be added to fim_run.sh to use or not use this feature.

nickchadwick-noaa:
Added a module to run at the end of `run_by_unit.sh` that will handle any whitelisting and post-processing of the output files at the HUC level.

## Additions

- Added `output_cleanup` module

## Changes

- Implemented the new `output_cleanup` module at the end of `run_by_unit.sh`
- Wired up new argument flags in `fim_run.sh` for managing whether or not a run is for production or Viz

Resolves #88 #94 and #111

### DIFF
--- a/fim_run.sh
+++ b/fim_run.sh
@@ -17,6 +17,10 @@ usage ()
     echo '  -j/--jobLimit   : max number of concurrent jobs to run. Default 1 job at time. 1 outputs'
     echo '                    stdout and stderr to terminal and logs. With >1 outputs progress and logs the rest'
     echo '  -o/--overwrite  : overwrite outputs if already exist'
+    echo '  -p/--production : only save final inundation outputs'
+    echo '  -w/--whitelist  : list of files to save in a production run in addition to final inundation outputs'
+    echo '                     ex: file1.tif,file2.json,file3.csv'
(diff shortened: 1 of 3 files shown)


nickchadwick-noaa:
Another thing to mention about this new feature. Since it is only deleting intermediate output files and creating new output files( for Viz ), this feature will not affect any previously computed metrics and will not cause any regressions to the fim_run process.

BradfordBates-NOAA (approved):
I tested both the -v and the -w flags. Both worked exactly as expected. It was observed that the rating_curves_12345678.json is a smaller file than the src.json. This is explained by the whitespace present in src.json. The two files are identical otherwise.

This enhancement marks a significant reduction in data footprint when the included features are used. Approved.

BradfordBates-NOAA:
I also confirmed that the -p flag results in the desired behavior. I recommend this be merged into dev.

Issue #160, opened 2020-12-14
[5pt] Fix no data value handling for ndv >0 cases


RyanSpies-NOAA:
Some elev_cm.tif files use >0 ndv. This leads to issues in the "Zero out negative values" step in run_by_unit.sh.

Issue #153, opened 2020-12-10
[3pt] Add masking for Great Lakes water area


RyanSpies-NOAA:
Mask REM for Great Lakes using the same approach we are using for ocean areas along coast.

PR #163, opened 2020-12-15
[1pt] PR: Fix for >0 no data values & masking Great Lakes water bodies

RyanSpies-NOAA:
Some of the raw NHDPlus DEMs in the Great Lakes region (e.g. 0402-0411) use large positive values to designate no data pixels (e.g. 65535). This causes issues in the gdal_calc formula used in the REM zeroed step. Incorporated previous fix for this issue from [commit 7e6339](https://github.com/NOAA-OWP/cahaba/commit/7e6339c27ef68b683b24ea8411c0016521aa29d1?branch=7e6339c27ef68b683b24ea8411c0016521aa29d1&diff=split). Future consideration: may want to set up a more robust catch in run_by_unit.sh or in the preprocessing workflow to reassign ndv to avoid potential issues with masking positive values.
Resolves issue #160  

Use a Great Lakes water body polygon to mask REM for HUC=04. Using gl_water_polygon.gpkg input. Applies the same methods as coastal/ocean masking. If statement checks if HUC=04 (Great Lakes) region and then assigns the landsea input file.
Resolves issue #153 

### run_by_unit.sh
- new if statement to check which HUC2 and then assign either the Great Lakes polygon or coastal/ocean polygon for masking (input_LANDSEA)

### snap_and_clip_to_nhd.py
- added gdal.clip step to trim the landsea water polygon to the wbd_buffer boundary. This helps reduce the workflow's data footprint and processing time.

## Changes

- modified the gdal_calc equations --> removed `+((A<=$ndv)*$ndv)` in the rem_zeroed.tif and rem_zeroed_mask.tif generation
- also modified a couple of print/echo statements to reflect recent changes

## Testing

1. Suggest running a Great Lakes region huc (e.g. 0408) 

## Notes

- None of these changes should have any impact on the BLE evaluation metrics

## Screenshots
ndv_problem
gl_mask_prob
gl_mask_prob2
final_results

### DIFF
--- a/lib/run_by_unit.sh
+++ b/lib/run_by_unit.sh
@@ -90,11 +90,6 @@ echo -e $startDiv"Get DEM Metadata $hucNumber"$stopDiv
 date -u
 Tstart
 read fsize ncols nrows ndv xmin ymin xmax ymax cellsize_resx cellsize_resy<<<$($libDir/getRasterInfoNative.py $outputHucDataDir/dem.tif)
-if [[ ${ndv%%.*} -ge 0 ]] ; then
-	echo -e "$hucNumber raster NDV is >= 0 --> resetting to -2147483648"
-	ndv=-2147483648
-fi
-Tcount
 
 ## RASTERIZE NLD MULTILINES ##
 echo -e $startDiv"Rasterize all NLD multilines using zelev vertices"$stopDiv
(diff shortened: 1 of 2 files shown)


BrianAvant-NOAA (approved):
Looks good. We should try to communicate the no data value issue to USGS to see if we can get that fixed on their end.

PR #200, opened 2021-01-06
hotfix to address AHPSs mapping errors

BrianAvant-NOAA:
Resolving errors that surfaced during AHPs library generation #201

## Changes

- checks dtype of hydroTable.csv columns to resolve errors caused in inundation.py when joining to flow forecast
- exits inundation.py when all hydrotable HydroIDs are lake features
- updates path to latest AHPs site layer


TrevorGrout-NOAA:
I approve this pull request. These fixes corrected some of the errors we were seeing with generation of static FIM libraries across the NWM domain, namely errors associated with leading zeros being removed, dtype issues associated with feature_id field and a graceful exit if hydroTable has no records if lake segments are removed.

TrevorGrout-NOAA (approved):

Issue #279, opened 2021-02-25
[8pt] Convert CatFIM pipeline to open source


BrianAvant-NOAA:
Convert ahps libraries pipeline to open source and integrate into the Cahaba repo.

- Convert arcpy dependent script that generates the CatFIM layer `mosaic_ahps_libraries.py`
   - [ ] additional files are used by that script (point files with attributes that are joined to the dissolved layer) and may need to be converted as well
   - [ ] The final layer needs to be dissolved by "ahps_lid;magnitude;version;huc"

- optimize `generate_categorical_fim.py` 
   - [ ] Add a function called `post_process_cat_fim_for_viz()` that calls `mosaic_ahps_libraries.py`
   - [ ] Add "version" (-v) as an arg to allow population of the version attribute
   - [ ] Final output must be feature class in geodatabase (using a template GDB)
   - [ ] Parallelize flood category processing

BradfordBates-NOAA:
Hey @BrianAvant-NOAA I don't see a reason to have `post_process_cat_fim_for_viz()` call `mosaic_ahps_libraries.py`. I think we can deprecate mosaic_ahps_libraries.py and put all of its logic (after converting to open source) into `post_process_cat_fim_for_viz()`. Thoughts?

BrianAvant-NOAA:
> 
> 
> Hey @BrianAvant-NOAA I don't see a reason to have `post_process_cat_fim_for_viz()` call `mosaic_ahps_libraries.py`. I think we can deprecate `mosaic_ahps_libraries.py` and put all of its logic (after converting to open source) into `post_process_cat_fim_for_viz()`. Thoughts?

That is more or less what I meant. I will reach out for feedback before I submit a PR.

PR #292, opened 2021-03-03
[1pt] PR: Convert CatFIM pipeline to open source

BrianAvant-NOAA:
Convert CatFIM pipeline to open source to resolve issue #279

## Changes

- added `VIZ_PROJECTION` to `shared_variables.py `
- added missing library referenced in `inundation.py`
- cleaned up and converted evaluation scripts in categorical FIM generation to open source
- removed `util` folders under `tools` directory


BradfordBates-NOAA:
Hey Brian, here are some things I noticed upon initial review:

- [ ] sys imports `sys.path.insert` need to account for Cahaba restructure
- [ ] PREP_PROJECTION is not in shared_variables.py anymore, fails to import
- [ ] I recommend maintaining [PEP8 styling](https://www.python.org/dev/peps/pep-0008/), with 2 returns before a function definition, also capitalized docstrings
- [ ] There are some variables used that are not defined, e.g. `log_file`, `log_dir` in the `reformat_inundation_maps` function

BradfordBates-NOAA:
I confirmed that the script ran all the way through. Nice!

I have some feedback regarding the fields that I did not mention in previous comments:
- [ ]  The final layer needs the following fields:

    -  `q`: Discharge for the respective flood category
    - `stage`: Stage for the respective flood category
    - `q_src`: Discharge source 
    - `stage_src`: Stage source
    - `q_uni`: Discharge units
    - `stage_uni`: Stage units
- [ ] Please round `q` and `stage` to the nearest decimal point

The values for these fields can be derived from the existing fields that are available post-join (e.g. `Q_act`, `S_min', etc.), however all fields need to essentially be "dissolved" into the fields in the list above. Here's a screenshot of the source data attribute table for the CatFIM service:

image

BrianAvant-NOAA:
> 
> 
> I confirmed that the script ran all the way through. Nice!
> 
> I have some feedback regarding the fields that I did not mention in previous comments:
> 
>     * [ ]   The final layer needs the following fields:
>       
>       * `q`: Discharge for the respective flood category
>       * `stage`: Stage for the respective flood category
>       * `q_src`: Discharge source
>       * `stage_src`: Stage source
>       * `q_uni`: Discharge units
>       * `stage_uni`: Stage units
> 
>     * [ ]  Please round `q` and `stage` to the nearest decimal point
> 
>     * [ ]  Please change `WRDS_time` to `wrds_time`
> 
> 
> The values for the fields listed above can be derived from the existing fields that are available post-join (e.g. `Q_act`, `S_min`, etc.), however all fields need to essentially be "dissolved" into the fields in the list above. Here's a screenshot of the source data attribute table for the CatFIM service:
> 
> image

Almost all of these are related to the input table I am joining to the spatial layer. Seems like those changes should be made there and not in this workflow. I'll get with @TrevorGrout-NOAA about resolving these.

BradfordBates-NOAA:
I think the fields in the list can only be created after the join because of the way the join table and the CatFIM library are formatted. We can chat more about it on a call though.

BrianAvant-NOAA:
Should be good to go. Trevor said he would deal with rounding in his feature branch.

BradfordBates-NOAA:
I ran it again and everything looks great, although we need to add `WFO` as an attribute. I've talked with @TrevorGrout-NOAA about this.

BradfordBates-NOAA (approved):
Approving. The `WFO` attribute will be represented after it is available in the necessary join table.

Issue #274, opened 2021-02-23
[8pt] Add Flood of Record to CatFIM (Vlab #88486)


BradfordBates-NOAA:
- [ ] Update WRDS API calls to include flood of record query
- [ ] Update generate_categorical_fim.py to produce mapping for flood of record magnitude

Issue #273, opened 2021-02-23
[8pt] Diagnose NWC HAND FIM Incorrect Rating Curve in Southeast Sites (Vlab #88478 and #88488)


BradfordBates-NOAA:
This is for FIM version 3.0.0.3 (which has been updated).

From Kelly Godsey:

The HAND FIMs for three sites in South Georgia (HRAG1, VDSG1, and QUTG1) are using rating curves without applying the correct adjustments to match the datum used by the NWS. These three sites are part of the Suwannee Basin and all stations within the Suwannee Basin changed to report water surface height above sea level in 2015. So for the HAND FIMs at these sites, referencing the USGS rating curve without applying the necessary correction, results in depiction of flooding from a flow that is unrealistic at each specified category. In the attached image for VDSG1, you can note a couple of things, 1) very little variation between the flood areas at the four flood categories, but more importantly, note the model specified flow of over 277kcfs. Simply not possible - and a product of not applying the correction for stage reporting to the rating curve. The NRLDB rating for these three sites would likely produce a much better result.

Additionally - there may be other errors throughout the Suwannee Basin as the NWS uses NAVD88 as the datum for all Suwannee sites in Florida, whereas the USGS still may be referencing NGVD29 as their primary datum. However, in Florida, the difference would not be as dramatic, as the Florida sites have always been on a sea level reporting system. The shift from NGVD29 to NAVD88 in this portion of Florida is generally between -0.6ft and -0.9 ft.

![image](https://user-images.githubusercontent.com/69594899/108878195-9f817280-75c5-11eb-93dc-dddb9e469e6d.png)

BradfordBates-NOAA:
More info from Kelly (88488): 

After seeing the rating curve issues at 3 South Georgia sites (ticket #88478), I went ahead and checked the entirety of the Suwannee basin (forecast points only) and there are several rating curve issues that are affecting the HAND FIMs to the point evaluations cannot be completed at most sites due to significant overflooding.

In Georgia - Sites HRAG1, VDSG1, QUTG1, and STNG1 all share the same issue explained in ticket #88478

In Florida, it was more than just a NGVD29 to NAVD88 conversion issue as some of the older stations are using separate gage datums, thus conversions to current NWS levels are more significant. All sites in the Suwannee Basin in Florida report water surface height above sea level in NAVD88.

WSPF1 - Mismatch between NGVD29 and NAVD88. NAVD88 is 0.83 ft lower at this gage
[...]

Issue #271, opened 2021-02-23
[8pt] Enable CatFIM at all AHPS sites via cahaba


TrevorGrout-NOAA:
Functionality to create flow files for CAT FIM at all ahps sites for CI/CD.

PR #298, opened 2021-03-10
[1pt] PR: Generate flow files for CatFIM

TrevorGrout-NOAA:
Addresses github #271 and github #274

Ability to generate flows for catfim. All external functions were placed in the "tools_shared_functions.py". This branch also utilizes .env files which contain sensitive paths/urls. This feature branch will search for all nws_lids in CONUS that have the attribute "rfc_forecast_point" as well as all nws_lids in HI/PR.  It intersects the nws_lid points with the FIM HUC 08 layer to create a directory of hucs and subdirectories of nws_lids. Within each subdirectory are subfolders of each threshold category (action, minor, moderate, major, and record) which are populated with flow files used for inundation mapping. Each flow file contains a list of NWM segments that are on the mainstem and a flow (in cms). Mainstem NWM segments are determined using a function using a multi-pass algorithm to get all NWM segments that are downstream of nws_lid sites. At the end of the script all info files are aggregated into a master info file which is placed in the parent workspace directory. An example call for this function is:

Currently it takes approximately 2.5 hours to run. This is quite slow compared to runtime of about 45 minutes on personal pc. To run the script you must checkout the "dev-catfim-flow" docker image as it contains additional required python packages.
`python3 generate_categorical_fim_flows.py -w /path/to/workspace -u 10 -d 10`

Some potential tickets that may need to be created:
There is some uncertainty on the CRS used by WRDS for the lat/lon locations. The script currently uses information that WRDS reports, however, an update may need to be implemented when CRS is sorted out with WRDS, especially if a new field is reported by WRDS.

Additionally, when writing the flow file, WRDS reports potentially two calculated flow sources: NRLDB and USGS. Currently, the USGS is given priority. It has been reported that for at least some sites (github #273 ), the NRLDB may be more accurate than the USGS. This may require a future enhancement.

Reasons why a site is not produced include:
It did not intersect with the FIM HUC 8 layer (outside the domain)
Outside the domain of the NWM (Alaska)
It does not have a name specified.
It does not have any flows specified for all thresholds.
No NWM segments


BradfordBates-NOAA:
This branch is now behind dev. Please update before review.

BrianAvant-NOAA:
@TrevorGrout-NOAA @BradfordBates-NOAA 

merged with FIM v3.0.8.2
updated `generate_categorical_fim.py`

Requested changes

- [ ] make sure that `nws_lid_attributes` table outputs HUC column as a string with leading zeroes
- [ ] run entire categorical fim process for `catfim_gaps_ms_c` outputs to test

TrevorGrout-NOAA:
I have updated the feature branch with dev. I have updated the feature branch to not strip off leading zeros in file outputs. Output spatial files are being projected based on the USGS supplied projection (which is currently mis-reported by WRDS) and then reprojecting everything to WGS 84.

It takes a couple hours to run this so I am currently running it to make sure it I didn't break anything, but I also propose that instead of having the reviewer rerun the entire process in order to merge, if my run is successful then we use that as evidence and I can point the reviewer to the outputs. As far as review, the main thing would be to have an external eye maybe take one site and compare it to WRDS and verify correct flow is being passed to flow file for a couple thresholds?

@BrianAvant-NOAA WRDS currently mis-reports the lat/lon datum. I have been in discussions with the WRDS developers and they have implemented a fix that won't be pushed until tonight. I propose running the 'official' catfim flows this evening/weekend and then running the generate_categorical_fim.py on Monday morning. 

I feel that if someone does a quick look at the flow files being created and maybe look at spatial output (?) that should be sufficient to merge this feature branch. When WRDS updates their API we can run the final CatFIM this evening.

TrevorGrout-NOAA:
@BradfordBates-NOAA @BrianAvant-NOAA 
Also, I do think a reviewer should at least start a run to make sure the .env file is working correctly, but not necessarily run the whole 2 hour process.

BradfordBates-NOAA:
I can do a test run. Can you please make the output spatial projection Web Mercator? You can import the projection info from utils (`VIZ_PROJECTION`).

BradfordBates-NOAA (approved):

BradfordBates-NOAA:
All in all, this looks really good. Here are a few comments:

- [ ] Ensure no more than 2 decimal places in `q` and `stage` fields in `catfim_library.shp`
- [ ] Sync the `mapped` column in `nws_lid_sites.shp` with the `viz` column in `catfim_library.shp` (a join should work here)
- [ ] In `nws_lid_sites.shp`, change values of `mapped` column to all lowercase, i.e. "yes", "no"

Issue #305, opened 2021-03-12
[8pt] Categorical FIM shapefile add attributes for mapped and reason


TrevorGrout-NOAA:
Categorical FIM currently produces two shapefiles. 
candidate_sites.shp -- This file contains all potential nws_lid sites
all_mapped_ahps.shp -- This file is all the nws_lid sites that actually had flow files generated.

Modify code so that candidate sites contains attributes: 
Mapped (True/False)
Notes (populated for each site and the contains the mapping status of each threshold for that site).

Issue #318, opened 2021-03-22
[5pt] Incorrect geometries in CatFIM merged layer


TrevorGrout-NOAA:
Additionally, there appears to be another issue with some sites/thresholds not appearing in correct location/having incorrect (and incomplete) geometry when compared to the grids. For example, id8s2 -record threshold-huc 10160009 and sfrw4-record-threshold-14040107. 

The CatFIM process was run 2 times and the same issue remained.

BradfordBates-NOAA:
@TrevorGrout-NOAA @BrianAvant-NOAA Is this still an issue?

Issue #317, opened 2021-03-22
[5pt] CatFIM raster to polygon errors


TrevorGrout-NOAA:
When generating CatFIM a warning is issued 

```
...geodataframe.py:432: FutureWarning: Assigning CRS to a GeoDataFrame without a geometry column is now deprecated and will not be supported in the future.
  return GeoDataFrame(rows, columns=columns, crs=crs)
```

Upon inspection this error occurs with site prdt2-minor threshold-huc 13040203. It appears that although the grid looks reasonable, when converting from grid to polygon it returns an empty geometry thus triggering the warning.  Upon review of the output CatFIM library shapefile, there appears to be another site that may have a corrupted geometry as well (hrdm8 - moderate threshold-huc10080016).

TrevorGrout-NOAA:
Here is some code to determine which geometries are invalid.
```
import geopandas as gpd
import pandas as pd
from shapely.geometry import Polygon

shp = r'/path/to/merged/shapefile
shp_df = gpd.read_file(shp)

invalid = pd.DataFrame()
for row in shp_df.itertuples():
    if not row.geometry.is_valid:
        rowdf = pd.DataFrame({'lid':[row.ahps_lid],'mag':[row.magnitude],'huc':[row.huc],'note':['invalid geometry']})
        invalid = invalid.append(rowdf)
```

PR #319, opened 2021-03-22
[3pt] PR: Dev Catfim spatial

TrevorGrout-NOAA:
Update to 3.0.10.0 to implement a few additional features and stability fixes.

Updated fixes to handle WRDS API output (with recent updates that were implemented).
Renamed scripts to better describe what they do (generate_categorical_fim_flows.py and generate_categorical_fim_mapping.py).
Added a wrapper script (generate_categorical_fim.py) that links the "flows.py" and "mapping.py" scripts.
Updated the status outputs of the nws_lid layer and saved it in the same directory as the merged catfim layer.
Additional minor stability fixes.

There are a couple residual bugs that need to be sorted out that appear to affect some sites (see issues #317 & #318).  Total runtime is quite lengthy and should be run overnight. An example run is as follows:

python3  generate_categorical_fim.py -f fim_3_0_9_0_ms_c -j 13

Will the reviewer please examine the output shapefiles in the "mapping" directory.  The nws_lid_sites shapefile has a "mapped" field and a "status" field. The "mapped" field indicates whether a site contains mapping in the catfim_library shapefile. If not, then the "status" field explains why:
Missing nwm segments --> No segments associated with site.
Missing threshold stages --> No stages supplied with site.
Missing calculated flows --> Stages were supplied but flows not available (e.g. exceeded rating curve).

If any site only has some thresholds that have flow files created then those are noted. The number of sites indicated as having mapping in the nws_lid_sites match the sites listed in the merged CatFIM spatial layer.  **The status field does NOT have information related to geometries are in the CatFIM spatial layer but may not be valid.**

This feature branch resolves tickets #305, #321, #322


BradfordBates-NOAA:
All in all, this looks good. Here are some comments:

- [ ] Sync `mapped` column with `viz` column (a join should do this)
- [ ] Ensure no more than 2 significant digits in `q` and `stage` fields
- [ ] Confirm lower case `yes` and `no` in `viz` and `mapped` fields

TrevorGrout-NOAA:
@BradfordBates-NOAA 

These should be linked as-is. The "viz" column is appropriately populated as "yes" for all records in the catfim_library.shp. If a threshold does not have spatial data it would not be in the catfim_library.shp.  The nws_lid_sites.shp layer lists all sites, and gives explanation for why a site may not be included in the catfim_library.shp. But all sites in the nws_lid_sites.shp that are designated as having mapping are also found in the catfim_library.shp. This was manually verified.

The catfim_library.shp file has no more than 2 decimal places  for "q" and "stage" attribute fields when viewed in ArcMap.  Can you describe where you saw the behavior you mention?

All "Yes","No" were converted to lower case

BradfordBates-NOAA:
Okay thanks for confirming the fields are synced and for changing to lowercase. Here's a screenshot of the catfim_library's attribute table, showing the significant digits in Q. Maybe it's just a Q thing?

image

TrevorGrout-NOAA:
This may be a difference between QGIS and ArcMap:
image

TrevorGrout-NOAA:
One last commit to ensure column naming consistency between catfim_library.shp and nws_lid_sites.shp

BradfordBates-NOAA:
@TrevorGrout-NOAA Please merge with latest dev and then we can merge this in.

TrevorGrout-NOAA:
@BradfordBates-NOAA I have updated this with recent changes to dev. Please advise what version of dev this will be and I will update changelog accordingly.

PR #320, opened 2021-03-23
[1pt] PR: Patch synthesize test cases

BradfordBates-NOAA:
Patches a bug where only previous versions could be synthesized in `synthesize_test_cases.py`


RyanSpies-NOAA (dismissed):
Reviewed code change and confirmed the newly generated master_metrics.csv contains feature branch metrics.

Issue #289, opened 2021-03-01
[5pt] Obtain USGS rating curve and convert to elevation


TrevorGrout-NOAA:
Convert USGS rating curve to elevation (NAVD). Subtask to #269

TrevorGrout-NOAA:
Started Feature Branch 'dev-site-rating-curve' to begin this work.
General tasks are to:
1*) Get all sites that rating curves are desired
2) If site has an NGVD datum convert to NAVD 
3) Retrieve rating curves for all sites (at this time focus on USGS rating curves)
4) Add an elevation column to the rating curve

*As subtasks coordinated with WRDS team to provide import gage information related to methods of datum collection and vertical accuracy of vertical datum. This will allow us to filter out sites that may not have high quality datum information.

PR #338, opened 2021-04-05
[1pt] PR: Get USGS rating curves (FB: dev-usgs-rating-curve)

TrevorGrout-NOAA:
This feature branch addresses #289. It will retrieve the USGS rating curve from the WRDS API and will perform a conversion of the rating curve from stage to elevation (in NAVD88). This is accomplished by adding the datum to the rating curve stages. If the datum is in NGVD29, the NOAA tidal api is used to convert the datum to NAVD88. If the "all" option is used, a successful run will query the WRDS API for all active usgs gages. These gages are then filtered to remove sites that may not be suitable for analysis (inaccurate datums, lat/lon coords etc). A geopackage containing these filtered sites as points is saved. Rating curves for sites in CONUS are retrieved and converted to NAVD88 elevation using the NOAA tidal API. The Tidal API call as currently constructed is not able to convert (NGVD29 to NAVD88) outside CONUS and as such these sites, although included in the spatial file, are not included in the CSV. 

## Additions
- New script: rating_curve_get_usgs_curves.py
- Creates 3 output files:
- 1) usgs_rating_curves.csv: A csv file that contains rating curves (including converted to NAVD88 elevation) for USGS gages in a format that is compatible with "rating_curve_comparisons.py"
- 2) log.csv: A log file that records status for each gage and includes error messages.
- 3) If "all" option is used, usgs_gages.gpkg:  This is a geospatial layer of all USGS gages that meet a predefined criteria (Datum accuracy, datum collection method, horizontal accuracy of coords, gage type and etc) already converted from differing source coordinate systems to the FIM default projection.  Additionally, a column is added which indicates whether a gage has been included in the usgs_rating_curves.csv. 
-  

## Removals

-

## Changes

- Modify usgs_gage_crosswalk.py to:
- 1) look for newly changed field "location_id" instead of "site_no". 
- 2) Filter out gages (using the usgs_gages.gpkg "curve" field) that do not have rating curves included in the usgs_rating_curves.csv
- Modify "rating_curve_comparison.py" to do a check on the age of the user specified "usgs_rating_curves.csv" and alert user to the age of the file and recommend updating if file is older the 30 days. 

## Testing

1. An example call for the script:
**python3 /foss_fim/tools/rating_curve_get_usgs_curves.py -l all -w /path/to/output/workspace**
In some instance the Tidal API seems to have errors if called to frequently. I have added a -t argument which can extend the rest between API calls. This will obviously lengthen the time it takes to run the script. Average run time could be up to a few hours.

Additionally, I verified that run_by_unit.sh runs with edits made and that rating_curve_comparison.py also runs with usgs_rating_curves.csv.


BrianAvant-NOAA (dismissed):

TrevorGrout-NOAA:
@BrianAvant-NOAA I updated the changelog. Please re-approve.
FYI, this PR does will not affect any FIM metrics.

Issue #330, opened 2021-03-29
[2pt] Attempt CONUS MS simulation with bathy feature branch


RyanSpies-NOAA:
Test the bathy rating curve adjustment feature branch with a mainstem CONUS run.

Issue #328, opened 2021-03-29
[2pt] Create user input flag on/off option for bathy estimation calculation (submit PR)


RyanSpies-NOAA:
Create option for user to toggle on/off bathy estimation rating curve component. This will allow me to submit a PR for the feature branch while it's still in prototype phase.

Issue #306, opened 2021-03-15
Agile Story: Test prototype solutions for representing channel bathymetry in synthetic rating curves


RyanSpies-NOAA:
Continuation of agile story from previous sprint (see card #282)

Explore solutions to account for known issue of missing channel bathymetry (FIM over-estimation at < minor flood conditions). The current solution incorporates regression-based channel bankfull geometry estimates to quantify the missing bathymetric component not represented in the DEM and resulting synthetic rating curve.

- [x]  [5pt] Develop SRC workaround solution for thalweg burn notch (Issue #309)
- [x] [3pt] Test that current feature branch solution is cable of running for large subset of HUCS (Issue #310)
- [x]  [3pt] Perform alpha evaluation for subset of HUCS (Issue #311)
- [x] [3pt] Create user input flag on/off option for bathy estimation calculation (submit PR) (Issue #328)
- [x] [5pt] Create slide deck content to document alpha eval performance
- [x] [2pt] Attempt CONUS MS simulation with bathy feature branch (Issue #330)

PR #340, opened 2021-04-07
[2pt] PR for bathy SRC prototype

RyanSpies-NOAA:
This code modification includes a prototype technique to estimate the missing bathymetric component in the HAND-derived synthetic rating curves. This new Bathymetric Adjusted Rating Curve (BARC) function built within the fim_run workflow will ingest bankfull geometry estimates input by the user to modify the cross section area used in the synthetic rating curve generation. 

Addresses tickets: #306, #328, #309, #311, & #330

## Additions

- bathy_rc_adjust.py: 
  - imports the existing synthetic rating curve table and the bankfull geometry estimates (topwidth and xs area per COMID)
  - performs all of the new synthetic rating curve calculations
  - attempts to flag issues with the thalweg-notch artifact

## Changes

- add_crosswalk.py: 
  - added streamorder to src_full_crosswalked.csv
  - calls the new bathy_rc_adjust.py if bathy env variable set to True and extent=MS
- run_by_unit.sh:
  - added new csv outputs for exploring BARC impacts
- params_template.env 
  - added new BARC function variables and on/off toggle variable
- params_calibrated.env 
  - added new BARC function variables and on/off toggle variable
- output_cleanup.py
  - added BARC csv files to production whitelist
- eval_plots.py
  - added additional AHPS eval sites to list of "bad_sites" due to issues with MS flowlines

## Testing

Alpha Evaluation performed for subset of 138 HUC8 sites
image
image
image
image

### Problem
image
image
image
### Objectives
image
### Approach
image
image
image
image
image


TrevorGrout-NOAA:
I was able to confirm that the feature branch runs without error for HUC 17090012. I was able to confirm that the rating curve is adjusted in the rating curve.  I inspected inundation output for huc 10190012 (FOMC2) and compared FIM 3090 ms c vs FB and found reduced mapping for both action and major stages with this new feature. CSI scores were approximately the same for major but considerably improved for the action stage (FB>3.0.9.0). 

After discussion with Ryan, I recommend a ticket be issued to find the optimal parameter values for this new enhancement. It would be nice if optimal data values could be determined prior to the next release of CatFIM.

Would it be possible to show metrics comparisons with NWS and USGS? They should be generated at the same time.  I am curious if the decrease in metrics is also found with NWS datasets?

TrevorGrout-NOAA:
I ran the plot scripts for USGS and NWS
image
image
I am noticing the NWS seems to have a more pronounced regression than USGS. But in both sets there is an improvement in the lower magnitude flows.

TrevorGrout-NOAA:
Please make sure the change log is updated and that your feature branch is up to date with dev.

RyanSpies-NOAA:
@BradfordBates-NOAA any guidance on the FIM version I should use for this PR? Looks like we are currently on FIM v3.0.14.0. I don't think this is a "breaking" change although it will require a new input file (bankfull geometry) in order to run.

BradfordBates-NOAA:
@RyanSpies-NOAA, this will be 3.0.15.0, as it includes a new feature enhancement and is not a patch.

NICE WORK!

BradfordBates-NOAA:
@TrevorGrout-NOAA Thanks for including the plot scripts. That's a huge help.

TrevorGrout-NOAA (dismissed):

RyanSpies-NOAA:
@TrevorGrout-NOAA changelog has been updated for final approval. thx

RyanSpies-NOAA:
HAND_channel_profile_bankfull_v2

Issue #238, opened 2021-01-28
Missing incoming FR stream segment


BrianAvant-NOAA:
Incoming stream is dropped during hydroconditioning of HUC 12020003

## Behavior

It looks like the issue occurs in the generation of either the flow direction or flow accumulation rasters.

NHDPlusIDs: 30000800167144, 30000800145028, 30000800213678, 30000800161730
Catchment HydroID: 14320083

## Screenshots

![image](https://user-images.githubusercontent.com/20129718/106204199-33f0e480-6182-11eb-8ffd-da64e7bd7430.png)

RyanSpies-NOAA:
Note: LID [BOTN4 ](https://water.weather.gov/ahps2/hydrograph.php?wfo=phi&gage=botn4)is currently not represented in the nhd_headwaters_adjusted_ms. BOTN4 exists in the inputs/nws_lid.gpkg and there is an active USGS gage site (#[01380450](https://waterdata.usgs.gov/nj/nwis/uv?site_no=01380450)) associated with BOTN4 so I believe the MS flowlines should exist here. 
![image](https://user-images.githubusercontent.com/69868854/115446776-4f95e500-a1dd-11eb-96ce-0df285b4cbc2.png)

PR #358, opened 2021-04-16
NHDPlus HR raster preprocessing

BrianAvant-NOAA:
Preprocess NHDPlus HR rasters for consistent projections, nodata values, and convert from cm to meters. Resolves issue #300 and vlab ticket #89057

Some of the NHDPlus HR elevation data uses a positive nodata value. This causes issues in FIM hydroconditioning and produces unnatural stream geometries. This branch enforces a negative nodata value and checks for consistent projections.

example of before and after nodata correction for HUC 05140101:
05140101_nodata_correction

## Additions

-`preprocess_rasters.py` reprojects raster, converts to meters, and updates nodata value to -9999

## Changes
- `burn_in_levees.py` replaces the `gdal_calc.py` command to resolve inconsistent outputs with burned in levee values
- cleaned up log messages from `bathy_rc_adjust.py` and `usgs_gage_crosswalk.py`
- outputs paths updated in `generate_categorical_fim_mapping.py`  and `generate_categorical_fim.py`

- `update_raster_profile`: cleans up raster crs, blocksize, nodata values, and converts elevation grids from cm to meters
hucs with updated nodata values:
`'08060100', '08060201', '08060202', '08060205', '08030100', '08030201', '08030202', '08030203', '08030204', '08030205', '08030206', '08030207', '04020101', '04020103', '04020104', '04020201', '08010100', '08010202', '08010204', '08010205', '08010208', '08010209', '08010210', '04030101', '04030104', '04030106', '04030107', '04030108', '04030110', '04030112', '04030201', '04030202', '04030204', '08050001', '08050002', '08050003', '05120101', '05120102', '05120103', '05120104', '05120105', '05120106', '05120107', '05120108', '05120109', '05120110', '05120111', '05120112', '05120113', '05120114', '05120115', '05120201', '05120202', '05120203', '05120204', '05120205', '05120206', '05120207', '05120208', '05120209', '04040002', '04040003', '04050001', '04050003', '04050004', '04050005', '04050006', '04050007', '04060101', '04060102', '04060103', '04060105', '04070007', '04080101', '04080201', '04080202', '04080203', '04080204', '04080205', '04080206', '04090001', '04090003', '04090004', '04100002', '04100003', '04100004', '04100005', '04100006', '04100007', '04100008', '04100009', '04100010', '04100011', '04100012', '04100013', '04270101', '04200001', '04110001', '04110002', '04110003', '04110004', '05140101'`

[...]

PR #371, opened 2021-04-28
Refactor NHDPlus HR preprocessing workflow

BrianAvant-NOAA:
Refactor NHDPlus HR preprocessing workflow. Resolves issue #238

## Changes

- NHD streams, NWM catchments, and headwaters input layers are consolidated for MS and FR. HUC8 intersections are now included in the input headwaters layer and there are attributes that define the type of each point. Both streams and headwater points have a `mainstem` attribute.

- `clip_vectors_to_wbd.py` checks stream segments that intersect with the WBD boundary and removes incoming streams from the selected layers. This is to prevent them from being routed out of the HUC during the flow direction calculation. 

- Some gaps/overlap in coverage exist at the confluences of HUC8s. This is not likely to be fixed. FIM4 may provide a solution to this issue if we process FIM on a reach scale instead of HUC8s.

## Screenshots

FIM v3.0.15.5 MS HUC 03150107
image

FIM v3.0.15.8 MS HUC 03150107
image

Fixes MS deviation
image

## Testing

**will add testing when it finishes running**


TrevorGrout-NOAA:
When calculating aggregate scores across all ahps sites (nws & usgs) there are some small fluctuations but overall very similar scores:

  | 3_0_15_3 | 3_0_15_8
-- | -- | --
Action | 0.450 | 0.447
Minor | 0.483 | 0.483
Moderate | 0.542 | 0.544
Major | 0.603 | 0.602

It has been observed that odd, and often unrealistic mapping can occur when a stream segment doesn't traverse the entire catchment. FIM 3.0.15.2 exhibits this:
image

Note how the Agreement raster indicates overprediction for catchments that are full traversed by the stream centerline. Note how there is underprediction for the upstream most basin.
image

This feature branch exhibits much more realistic catchments and mapping:
image
The overprediction noted in 3.0.15.2 on the downstream catchments is now evident throughout.
image

Despite this site have reduced CSI, I would argue the output is more realistic and consistent. Altering Manning "n" would likely produce more agreeable results.

TrevorGrout-NOAA:
I was able to check out the feature branch (dev-local-headwaters) and successfully run HUC 03150107 and verified that the catchments appear as indicated. I also inspected IOWI4 and agree that this feature branch, although a reduction in CSI, produces more consistent and realistic results by changing how the catchments are made.  Additionally, when viewing the aggregate CSI of all AHPS sites (NWS & USGS) they are only vary by up to 0.3% between 3.0.15.3 vs this feature branch.

src/agreedem.py
```diff
@@ -45,8 +45,6 @@ def agreedem(rivers_raster, dem, output_raster, workspace, grass_workspace, buff
     # Import dem layer and river layer and get dem profile.
     elev = rasterio.open(dem)
     dem_profile = elev.profile
-    if elev.nodata == 0.0:
-        dem_profile.update(nodata = -999)
```
TrevorGrout-NOAA, line 48: Brian, Can you please elaborate why this change was added? Was this an oversight from earlier that is also fixed?

BrianAvant-NOAA: The changes in PR #358 make this code redundant.

TrevorGrout-NOAA (approved):

TrevorGrout-NOAA:
@Brian, please make sure this FB is up to date with dev prior to merging.

Issue #374, opened 2021-05-04
[3pt] Resolve FIM v3.0.15.9 MS non-zero exit codes


BrianAvant-NOAA:
Resolve FIM v3.0.15.9 MS non-zero exit codes related to Great Lakes coastlines.

## Current behavior
HUCs `04060101` and `04040002` are throwing errors in `run_by_unit.sh`  for FIM v3.0.15.9 MS. The issue is the Great Lakes coastline segments in the NHDPlus HR input data. Attributes in the dataset are not consistent enough to filter out coastlines so they will need to be masked in the FIM workflow. 

## Screenshots
![image](https://user-images.githubusercontent.com/20129718/117054994-4b3ef100-ace0-11eb-833c-1c46aa3b114f.png)

PR #375, opened 2021-05-04
PR: Remove Great Lakes from wbd buffer

BrianAvant-NOAA:
Remove Great Lakes from wbd buffer. Resolves issue #374

## Changes

-The `gl_water_polygons.gpkg` layer is used to mask out Great Lakes boundaries and remove NHDPlus HR coastline segments. These segments are causing issues later in `run_by_unit.sh` and unnecessarily increasing total processing time.

The work done on this feature branch is mostly recycled from @RyanSpies-NOAA's feature branch `dev-ocean-mask-dem`.

## Testing

Resolves non-zero exit codes for MS Great Lakes HUCs. No errors with FR/MS non-Great Lakes HUCS

## Screenshots
FIM v3.0.15.9 MS streams for HUC 04040002
image

MS streams and catchments for HUC 04040002 with hotfix
image

## Notes

Geopandas (Shapely) buffer doesn't have the same precision as GDAL so there is a slight difference in the WBD buffer layer geometries. The buffer is set to 5000 m and should be well beyond the HUC extent and I changed the Shapely default buffer resolution so the difference is negligible though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 Remove Great Lakes coastlines from WBD buffer.
 
-## Additions
+## Changes
 - gl_water_polygons.gpkg layer is used to mask out Great Lakes boundaries and remove NHDPlus HR coastline segments.
 
 <br/><br/>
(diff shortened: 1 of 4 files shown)


RyanSpies-NOAA (approved):
Tested this feature branch on huc 04080104. Confirmed that this runs for both MS & FR and produces desired results (removes NHDPlus coastline segments and improves fim_run runtime). 
image

Issue #436, opened 2021-07-19
Mechanism to automatically adjust synthetic rating curves given field feedback


BradfordBates-NOAA:
Proposed feature would ingest spatial points depicting known FIM boundary and use point locations to modify synthetic rating curves for relevant catchments to yield better matching inundation extents. A dummy spatial file may be used for testing.

The fields `HydraulicRadius (m)`, `WetArea (m2)`, `SLOPE`, `ManningN` will be used to programmatically update the SRC json file, so these three variables will need to be added to `Hyrotable.csv`. This can be done in `add_crosswalk.py` (line 223).

- [x] Write function to extract HAND and hydroid at any given point, as determined by a shapefile (#437)
- [x] Determine the median HAND value for each relevant Hydroid (#437)
- [x] Use median HAND value, flow, and hydroid and read in wetted area, hydraulic radius, and feature_id to perform the SRC adjustment (#)
- [x] Copy default flow and N columns to new columns and give new name, overwrite official columns with the new calculated values, also add `last_updated` and `submitter` columns

BradfordBates-NOAA:
Recommended Hydrotable 
![image](https://user-images.githubusercontent.com/69594899/127371072-8cf3b19b-732e-448a-a522-faadafea4b60.png)

BradfordBates-NOAA:
```python foss_fim/tools/adjust_rc_with_feedback.py -p /data/temp/water_edge_rc_adjust/known_water_edge.shp -d /data/outputs/point_adjust/aggregate_fim_outputs/ -w /data/temp/water_edge_rc_adjust/WBD_National_copy.gpkg```

PR #462, opened 2021-09-13
[2pt] PR: adjust SRC with user observed feedback points

RyanSpies-NOAA:
This new workflow ingests FIM point observations from users and “corrects” the synthetic rating curves to produce the desired FIM extent at locations where feedback is available (locally calibrated FIM). This approach is similar to the "pin to flood" methodology but differs in that this workflow aims to adjust the entire rating curve. Addresses #436 and #450 

## Changes

- `add_crosswalk.py`: added 'NextDownID' and 'order_' attributes to the exported hydroTable.csv. This will potentially be used in future enhancements to extend SRC changes to upstream/downstream catchments.
- `adjust_rc_with_feedback.py`: added a new workflow to perform the SRC modifications (revised discharge) using the existing HAND geometry variables combined with the user provided point location flow and stage data.
- `inundation_wrapper_custom_flow.py`: updated code to allow for huc6 processing to generate custom inundation outputs. 

## Testing

- To date all testing and development has been performed on huc6 aggregated outputs for 120401, 120903, and 102701

## Screenshots
image
Example showing two data points of "observed" stage & flow plotted on the raw HAND-derived rating curve

image
Example of the original and revised FIM extents using the observation point data

## Notes

Code methodology for adjust_rc_with_feedback.py:

1. Loop and find the corresponding hydroids in the Hydrotable
2. Grab slope, wetted area, hydraulic radius, and feature_id that correspond with the matching hydroids and HAND value for the nearest stage
3. Calculate new column for new roughness using the above info
4. If multiple flows exist per hydroid, aggregate the resulting Manning Ns → use median. If range of resulting Manning Ns is high, notify human (put this in a log file?)
6. Update hydrotable.csv
7. Copy default flow and N columns to new columns with "_default" in the field name
8. Overwrite the official flow and N columns with the new calculated values
9.  Add last_updated column with timestamp where values were changed
10. Update JSON file
11. Update SRC for catchments that have the same feature_id
12. If these catchments already have known data, then let it use those. If not, use new calculated Ns.

## Next Steps

- Test and evaluate this workflow with additional (and larger) observation point data


BradfordBates-NOAA (approved):
Nice work!

frsalas-noaa:
Does this commit define how far up and downstream the adjustment is applied or is that separate?

RyanSpies-NOAA:
Currently, the modifications to the SRC's are restricted to the hydroid
catchments that contain user-provided calibration points. I then do a
second pass to apply the same adjustments (i.e. modified manning's n) to
any hydroids that share a featureid. In other words, the changes are
limited to the featureid catchment where we have observation points. We are
not applying changes upstream or downstream, but I have the routing
attributes readily available in the hydroTable if we decide to try that.

frsalas-noaa:
Ok thanks. Please add a card in backlog to implement modification downstream. TBD on methodology for that but I understand some ideas have been discussed.

RyanSpies-NOAA:
Follow-on card #463

PR #472, opened 2021-10-26
[3pt] PR: Dev input validation

RobHanna-NOAA:
## Primary changes:
  - Input validation added to fim_run.sh to check all command line inputs for errors. 
  - Added more robust error handling and improved error messaging details so the user knows how to correct mistakes. 
  - Added a "step" system, so users can jump into any significant portion of run_by_unit.sh (via fim_run.sh). Aka.. They don't have to run the entire fim_run to test out one part such as burning in leeves for example. Note: The step system assumes that correct output files already exist if starting at later steps.
  - Added modularized validation functions that can be plugged into other bash files later. Most of the validation functions were replicated in python (but not 100% finished) so they can be used to validate input in other python files.
  - Most files were upgraded for improved in-line commenting and formatting more compliant to PEP-8 standards. Upgrades to ease of readability included when possible without major future merge conflict risks.

## Additions

- `src/validate_fim_run_args.sh`: Most input validation for all incoming parameters in fim_run is done here. Some of the incoming parameters have been changed for easier use down the line. ie) The input parameter -u (hucList) allows either a set of HUCs comma delimited or a path to a file that contains a list of HUCs. Previously, other bash and python files continued to work with either of the two possibilities. Now, working with check_input_hucs.py, ensure that all remaining processing files are always working with a list (collection) of HUCs no matter how they came in via the arg parameters. Also, adjusted some rules to allow flexibility for things like the -e (extent) can now accept MS or FR in any case combination. 
- `src/validate_fim_run_args.py`: An incomplete version of the bash version (validate_fim_run_args.sh). With a little updating, it will be available if/when all files are converted to python. Note: This file was initially created as the validation tasks were being completed in this python file. It was discovered that python struggles to send the complex messages back to Bash. The task was restarted using bash (.sh) files and it was decided to keep this python version for later. 

## Removals

- None

## Changes

* `fim_run.sh`: changes including:
    * Added input arg switches for `ssn` (Step Start Number), `sen` (Step End Number), and `api`. The `api` switch is not a published switch. It suppresses/simplifies some error messages when the fim_run.sh is not being called by a user via a terminal server (aka: simplified messages when fim_run.sh is being called by the "API" tool.)
    * Changed all incoming args to be global variables so other python and bash files can read/write to those variables (ie: the new validation tools and run_by_unit.sh).
    * Bash version of Try/Catch (called "Trap") added to handle errors more gracefully and be more detailed to end users (or tools) on what errors may have occurred. 

* `src/run_by_unit.sh`: changes include
    * Adding a step system. Each major part where another python files were being called to handle a task, became a "step". Each step is now numbered  and labelled as such and outputs adjusted accordingly. The step system allows (with command line args), to start and end at any step. ie) you can set the ssn (Step Start Number) at step 5 and run just that step 5; or  you can start at step 16 and run to the end; or run all steps up to step 34. In this version there are 41 steps identified.
    
* `src/bash_functions.env`: Add functions to process error checking: "Check_IsInteger", "Check_File_Folder_Name_Characters" and "Check_Valid_File_Name_Pattern" added.

* `check_huc_inputs.py`: changes include:
    * Major changes to improve validations and add new validations.
    * Ensuring that all output is now a list of HUCS. When a file path to a file of HUCs are loaded, it extracts all HUCs validates them, and outputs the list. 
    * Added Try/Catches to improve error handling and manage messages being passed back to other Bash files. Note: Messaging is still designed primarily to respond to Bash files and not python files. 
    
* `r_grow_distance.py`: Adjusted a few places where outputs were missing "overwrite" flags. This was required for the step system to work.

* `src/utils/shared_functions.py`: New functions added for input validations. Note: Also part of the "work in progress" attempt to add input validation via python versus Bash. Retained functions for later use and may require more testing.

* `src\output_cleanup.py`: Minor fixes around parallelization and cleaning up directories.

## Notes
* When using multiple HUCs for the -u args, it will work fine if the multiple HUCs just fine when in quotes. If a person forgets the quotes, it will only honor the first HUC and ignore the rest. We were unable to prevent or catch this. It appears this weirdness existed on previous versions.

## Testing

Over 50 test conditions were run covering a wide range of scenarios for input args and error handling such as:
* Invalid characters for some input args.
* Files that didn't exist or invalid pathing.
* Invalid values such not using MS or FR for the -e (extent).
* Invalid numbers when numbers expected.
* Handling spaces inside quotes for things like a list of HUCS in the -u (huclist).
* And many more.
* Lots of testing with combinations of HUCs, HUC files and whitelist files.
* Ran the regression test and confirmed that there was no impact on metrics.

## Todos

- If/when fim_run.sh, fun_by_unit.sh and others are switched from Bash to Python, the new "validation" files and tools will need to be completed and tested. Estimated to be 80 to 90% complete. Some partial python validation has been left and commented out in fim_run.sh for future changes.


BradfordBates-NOAA:
This PR is closed because future argument validation code will be added after the codebase has replaced Bash with Python.

PR #493, opened 2021-12-10
[1pt] PR: Adds ability to create a GPKG with joined rating curve comparison statistics

CarsonPruitt-NOAA:
This merge adds the functionality to create a geopackage with a point layer whose attribute table includes statistical comparison between SRCs and USGS rating curves.

## Changes

- `/tools/rating_curve_comparison.py`: 
    - updated recurrence intervals to reflect newest 17C run: 2, 5, 10, 25, 50, 100
    - added `create_static_gpkg()` function which is called when the optional `--stat_gages` option is passed and will create a GPKG that joins a USGS gage dataset with the Sierra Test statistics
- `/tools/rating_curve_get_usgs_gages.py`: bug fixes and added simple progress indicator for gage requests and metadata processing
- `/tools/tools_shared_functions.py`: added `requests_get_with_retry()` function that is a wrapper for the requests to API calls that will retry and wait if the response is throttled

## Testing

To test, you can look at the attribute table for `/tools/sierra_test/testing_versions/fim_3_0_24_2/usgs_gages_stats.gpkg` to confirm that the statistics have joined successfully with the gage dataset. There is a PNG with the same name that includes summary plots for each statistic.

You can also run the Sierra Test using the argument `-pnts /inputs/usgs_gages/usgs_gages.gpkg usgs_gages.stats.gpkg`. The `-pnts` accepts 2 arguments: 1) input USGS gage geopackage and 2) the output geopackage name, which will be written in the same output directory specified in the `-output_dir` argument.

usgs_gages_stats gpkg_summary_plots

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,12 +1,25 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v3.0.24.5 - 2021-12-10 - [PR #493](https://github.com/NOAA-OWP/cahaba/pull/493)
+
+This merge adds the functionality to create a geopackage with a point layer whose attribute table includes statistical comparison between SRCs and USGS rating curves.
+
+## Changes
+- `/tools/rating_curve_comparison.py`: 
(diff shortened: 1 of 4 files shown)


RyanSpies-NOAA (approved):
Approved.
Tested code on subset of hucs and confirmed metrics in gpkg and csv outputs

Issue #502, opened 2022-01-12
[5pt] Address usgs_gage_crosswalk.py artifact with attributing USGS gage to multiple hucs


RyanSpies-NOAA:
Attached screenshot shows example of a USGS gage that is associated to two different HUCS when running usgs_gage_crosswalk.py. UGSG gage 01357500 falls with the boundaries of HUC 02020004 but is also inside the wbd_buffered boundary for huc 02020003 (orange polygon). This causes duplicate gage location entries with different hydroid HAND elevation entries in the output usgs_elev_table.csv files.
![image](https://user-images.githubusercontent.com/69868854/149191352-5ca0ede6-ebd9-4f8d-ad5e-6648ee7dc2ed.png)

Issue #498, opened 2021-12-28
[5pt] Modify feature_id attribute extraction for usgs_gage_crosswalk.py


RyanSpies-NOAA:
usgs_gage_crosswalk.py currently crosswalks USGS gage points to the hydrofabric "feature_id" using a spatial join with the hydroid catchment polygon layer. This can result in attributing the wrong feature_id when the gage point resides near the boundary of hydroids. Suggest using the USGS gage snapped to the channel/thalweg before performing the sjoin to help avoid this issue.

CarsonPruitt-NOAA:
The `feature_id` mismatch has been traced back to an out-of-date AHPS site dataset. Catchments that have AHPS sites will use its attribute table in the crosswalk and will produce erroneous `HydroID` -- `feature_id` pairs, which will be passed into the `hydroTable.csv` output.

PR #503, opened 2022-01-12
[1pt] PR: refinements to usgs_gage_crosswalk.py

RyanSpies-NOAA:
This pull request adds an additional feature_id attribute associated with USGS gage locations (feature_id obtained from the WRDS API) output in the `usgs_gage_crosswalk.py`. There is also a new USGS gage query to check that gage HUC8 matches the FIM HUC when performing the gage-huc spatial query (avoids duplicating gages in multiple HUCs). Address #498 #502

## Changes

- `src/run_by_unit.sh`: updated to include the HUC8 id as an input argument to `usgs_gage_crosswalk.py` when running `fim_run.sh`
- `usgs_gage_crosswalk.py`: updated to include both `feature_id` (HAND-FIM crosswalked feature_id) and `feature_id_wrds` (feature_id assoicated with USGS gage location obtained from WRDS API). Also included the `gage_distance_to_line` attribute to potentially identify issues with correlating gage locations to the nearest thalweg pixel.

## Screenshots
Example of USGS gage that was previously associated to two different HUCS. UGSG gage 01357500 falls with the boundaries of HUC 02020004 but is also inside the wbd_buffered boundary for huc 02020003 (orange polygon). This pull request uses the HUC8 attribute within the USGS gage layer to perform a HUC query within 'usgs_gage_crosswalk.py`. The final result in gage 01357500 is now associated with huc 02020004 and not huc 02020003.
image

Example of a new usgs_elev_table.csv for huc 02020003 with the new attributes highlighted in yellow
image

## Notes

- Check with team to determine if there are any pitfalls with this solution for ensuring gage is only associated with one HUC

## Todos

- Need to apply these changes to a full BED run for MS and FR

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,17 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v3.0.24.8 - 2022-01-12 - [PR #503](https://github.com/NOAA-OWP/cahaba/pull/503)
+
+This pull request adds an additional feature_id attribute associated with USGS gage locations (feature_id obtained from the WRDS API) output in the `usgs_gage_crosswalk.py`. There is also a new USGS gage query to ensure gage's HUC8 matches the FIM HUC when performing the gage-huc spatial query (avoids duplicating gages in multiple HUCs).
+
+## Changes
+- `src/run_by_unit.sh`: updated to include the HUC8 id as an input argument to `usgs_gage_crosswalk.py` when running `fim_run.sh`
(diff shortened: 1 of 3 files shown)


CarsonPruitt-NOAA (approved):
Tested output on 4 HUCs. Code and outputs look good.

BradfordBates-NOAA:
@RyanSpies-NOAA Has this been tested at the BED scale? I see your comment about needing to apply these changes to a full BED run for MS and FR.

RyanSpies-NOAA:
@BradfordBates-NOAA I would like to kickoff a full MS BED run as soon as the API is updated to send outputs to the new shared dir.

BradfordBates-NOAA:
@nickchadwick-noaa Are we still waiting for the machine(s) to come back online?

RyanSpies-NOAA:
@BradfordBates-NOAA MS BED job is currently running on the API. I will check the results first thing tomorrow.

PR #515, opened 2022-01-28
[1pt] PR: Hotfix for usgs_gage_crosswalk.py error in Hawaiian HUCs

CarsonPruitt-NOAA:
Hotfix for usgs_gage_crosswalk.py error in Hawaiian HUCs `20020000` & `20070000`

## Changes

- `src/usgs_gage_crosswalk.py`: Use `pygeos.coordinates.get_coordinates()` to get the snapped coordinates instead of converting to shapley objects to get the coordinates. Using `pygoes` ensures that any Z-values are not included in the coordinates.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,17 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+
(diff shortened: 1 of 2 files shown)


BradfordBates-NOAA (approved):
The FR and MS BED runs finished without error, so I'll approve. But please update the Changelog @CarsonPruitt-NOAA, with target version being 3.0.24.14

BradfordBates-NOAA (approved):

PR #525, opened 2022-02-01
[1pt] PR: Adds basic unit test capabilities

RobHanna-NOAA:
This is the start of a very basic unit test system. I can also be used to assist in development efforts. Over time, more python files can have unit tests files created for them. Additional unit tests within unit test files will likely also be add (especially for "fail" conditions where one wants to ensure that bad data is being handled correctly in the code.

## Additions

- A new unit_tests folder which has the following:
   - clip_vectors_to_wbd_params.json:  A set of default "happy path" values that are expected to pass validation for the   clip_vectors_to_wbd.py -> clip_vectors_to_wbd (function). 
   - clip_vectors_to_wbd_unittests.py: A unit test file for src/clip_vectors_to_wbd.py. This file has only one test in it for happy path, but more are expected as "fail" test conditions are added later.
   - README.md:  Some information about how to create unit tests and how to use them.
   - unit_tests_utils.py: A python file where methods that are common to all tests can be placed. At this point, only one is included to create consistancy in unit path output print headers.
   - gms/derive_level_paths_params.json: A set of default "happy path" values that are expected to pass validation for the   derive_level_paths_params.py -> Derive_level_paths (function).  Note: the original method started with an upper case "D", so it used here as well. 

   -  gms/derive_level_paths_unittests.py: A unit test file for src/derive_level_paths.py. This file has only one test in it for happy path, but more are expected as "fail" test conditions are added later. Also, the "happy path (success)" function is not complete. it currently is evaluating the output only. 

## Notes

This is a very basic "starting" system. Even the two included unit test files are incomplete, but they have value even now for development purposes. This will hopefully create some precedence's that evolve. Eventually, the unit test system can become a tool that is triggered based on git checkin's to validate that a change has not broken other code functionality in other places.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,7 +4,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ## v4.0.1.0 - 2022-02-02 - [PR #525](https://github.com/NOAA-OWP/cahaba/pull/525)
 
-The addition of a very simple and evolving unit test system which has two unit tests against two py files.  This will set a precendence and will grow over time to become a full automated, possibly git checkin triggered, unit test system. The embedded README.md has more details of what we currently have, how to use it, how to add new unit tests, and expected future enhancements.
+The addition of a very simple and evolving unit test system which has two unit tests against two py files.  This will set a precendence and will grow over time and may be automated, possibly during git check-in triggered. The embedded README.md has more details of what we currently have, how to use it, how to add new unit tests, and expected future enhancements.
 
 ## Additions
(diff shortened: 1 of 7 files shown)


BradfordBates-NOAA:
@RobHanna-NOAA Please update the Changelog. I assume these additions have no impact on the GMS model, but are simply additional tools?

RobHanna-NOAA:
Correct, it does not affect functionality at all. Just a simple start to a unit test system but can be used for development purposes as well.

Change Log now updated.

BradfordBates-NOAA (approved):

PR #545, opened 2022-02-27
fixes #544: split_flows.py: tiny split reaches caused by mismat…

### LINKED ISSUE #544: split_flows.py: tiny split reaches caused by mismatch of flowline and lake polygon
DEM-derived flowlines may not align perfectly within a lake polygon, leading to tiny short split model reaches and corresponding catchments that look weird and may be hydrologically difficult to handle. A new algorithm is developed and will be submitted as a pull request.

## Current behavior
DEM-derived flowlines are overlapped with WBD8 boundary first to split boundary flowlines into inside and outside portions. Similar overlapping is applied on lakes. This is done by a simply GeoPandas overlay operation with `union` method. In GeoPandas 0.9.0, this overlay gives out warning on dropped splits outside of lake polygons, but harmless.

However, the following case on HUC8 12090205 (Tx Austin) shows such handling creates unexpected splits, illustrated below. Since the DEM-derived flowline does not fall entirely into the lake polygon, instead, it goes in and out of the lake several times, creating tiny splits as short as 0.17m.

image

image

## Expected behavior
The designer of this code can confirm, I think the intended flow splitting logic is to split a flowline that overlaps a lake into at most three parts: the one leads to the inlet of the lake, the one within the lake boundary, and the one coming out of the outlet. Given this assumption, I coded a new algorithm to split the flows when they overlap with lakes. The code handles the general case of multiple flowlines overlapping multiple lake polygons as long as no two lake polygons overlap.

The flow splitting after the fix:

image

Also, in the fix, another minor bug is fixed. Overlapping flowlines with HUC8 boundaries may also create empty geometries (HUC8 12100202). So empty geometries need to be removed after HUC8 overlapping, too.

## Steps to replicate behavior (include URLs)

The issue was discovered when I was producing HAND on 3m DEM for Texas.

hohe12ly:
the new algorithm uses GeoPandas overlay with `identity` method to create splits. then these splits are ordered by lake IDs. a subsequence that goes in and out of a lake multiple times is modified to have at most three parts: the one going into the late, the one in the lake, and the one out of the lake. the new algorithm returns such splits as a GeoDataFrame for the next step 1.5km length-based splitting. That length-based splitting is unchanged.

Also, added one step after HUC8 overlay to remove empty geometries.

### DIFF
--- a/src/split_flows.py
+++ b/src/split_flows.py
@@ -24,7 +24,242 @@ from collections import OrderedDict
 import build_stream_traversal
 from utils.shared_functions import getDriver, mem_profile
 from utils.shared_variables import FIM_ID
+import shapely # ornl
+import shapely.geometry  # ornl
+import rtree  # ornl
 
+# ornl
+def connect_seg_by_lake(lakeseq0):
+    '''
+    lakeseq0 is indexed by flowlines and the value of each element is
+    associated lake id. it is likely that a stream line goes in and out 
+    of a lake polygon multiple times, which produces a subsequence like
+    0,a,a,0,0,a,0,a,0. The intention of lake overlay is to split an
+    overlapped stream line into at most three parts: the part to the 
+    inlet to the lake, the part in the lake, and the part from the outlet
+    of the lake. this function converts the above subsequence to 
+    0,a,a,a,a,a,a,a,0 so as to label lake id to consecutive segments 
+    that belong to that lake.
+    observations: let's add a virtual lake id to the beginning -1
+    1. any lake seg begins with that lake id. no zeros in front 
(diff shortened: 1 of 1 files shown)


BradfordBates-NOAA:
@BrianAvant-NOAA Please check this out if you are interested

frsalas-noaa:
We should also note this for FIM v4.X.

RobHanna-NOAA:
Should we do it now while it is fresh in our heads?  It will certainly trip someone up someday. I can build unit_tests for each of those methods to validate the fix. The fix is quick, the fix validation is going to take a couple of days. Thoughts?

PR #548, opened 2022-03-01
[2pts] PR: New Feature to remove stream orders 1 and 2 if desired via command arg

RobHanna-NOAA:
By optionally including a "-s" command arg into `gms_run_unit.sh`, you can block loading any nwm stream records where the stream orders are 1 or 2.  By not including the "-s" arg, all stream orders will be loaded as gms was originally designed.

## Additions

- None

## Removals

- None

## Changes

- `gms_run_unit.sh`:  Addition of an optional -s switch, which if included will block GMS from loading nwm streams with stream orders of 1 and 2.
- `src/gms`
    - `run_by_unit.sh`:  Forwarding of optional new -s switch (dropLowStreamOrders) to 'derive_level_paths.py'.
    - `derive_level_paths.py`:  Added optional -s (drop_low_stream_orders variable). When not submitted, it defaults to "False" meaning all stream orders will be included. When "True", this flag is forwarded to 'stream_branches.py' to filter the streams.
    - `stream_branches.py`:  Added an optional argument to the `from_file` method called drop_low_stream_orders. When this is set to True, and the file being loaded into a GeoDataframe has the name "nwm_subset_streams" in it, it will filter out records with stream orders of 1 and 2.
- `unit_tests/gms`
   - derive_level_paths_params.json`: Added new drop_low_stream_orders input variable.
   - `derive_level_paths_unittests.py`:  
       - Upgraded: `test_Derive_level_paths_success` to include new param and also add more tests to validate success
               such as validating the correct record count return, and validating that output files were added to the file
               system as expected.
       - Added: two new unit tests methods:
             - One  for verifying that when the drop_low_stream_orders is not submitted, the system defaults to "False", aka.. no 
             filtering.
            - One for verifying that when submitting drop_low_stream_orders as true, that the number of resulting dissolved stream orders is reduced to the correct number.

## Testing

- Tested `gms_run_unit.sh` with the new -s (dropStreamOrderparam) command line argument being included (meaning drop the stream orders 1 and 2). Also tested with the '-s' arg not being submitted in command arguments, which resulted with no filtering took place as expected. For both scenarios, the files of nwm_subset_streams.gpkg, nwm_subset_streams_levelPaths.gpkg, nwm_subset_streams_levelPaths_dissolved.gpkg were loading into QGIS to validate record counts. This include manually filtering out stream orders against the full nwm_subset_streams in QGIS to validate new record counts.

- The adjustments to the 'derive_level_paths_unittests.py` inherently validate some inputs and outputs. All tests in this unit test file were successful, based on its twin "params.json" file.

### DIFF
--- a/src/gms/stream_branches.py
+++ b/src/gms/stream_branches.py
@@ -19,6 +19,20 @@ from scipy.stats import mode
 
 class StreamNetwork(gpd.GeoDataFrame):
 
+    '''
+    Notes:
+        - Many of the methods support two attributes called branch_id_attribute and values_excluded.
+          This can be used to filter out records. 
+            ie) When calling the nwm_subset_streams.gpkg, you can filter some records like this:
+                StreamNetwork.from_file(filename=outputs/<some huc>/nwm_subset_streams.gpkg,
+                                                 branch_id_attribute="order_",
+                                                 values_excluded=[1,2]
+                (which means drop all records that have an order_ of 1 or 2.
+                
+        - Note: from_file is using its branch_id_attribute and values_excluded as intended but other
+             methods may not be incomplete and not filtering as expected.
+    '''
+    
     geom_name = 'geometry' # geometry attribute name
     branch_id_attribute = None # branch id attribute name
     values_excluded = None
@@ -39,8 +53,8 @@ class StreamNetwork(gpd.GeoDataFrame):
 
 
     @classmethod
-    def from_file(cls, filename, branch_id_attribute=None,values_excluded=None, 
-                  attribute_excluded=None, verbose=False, drop_low_stream_orders=False, *args, **kwargs):
+    def from_file(cls, filename, branch_id_attribute=None, values_excluded=None, 
+                  attribute_excluded=None, verbose=False, *args, **kwargs):
 
         """ loads stream network from file to streamnetwork geopandas """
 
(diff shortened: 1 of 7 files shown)


RobHanna-NOAA:
Note: Branch can be deleted once approved.

RobHanna-NOAA:
Note: The second commit corrected incomplete functionality for the stream_branches.from_file. It already had params being passed into the method for branch_id_attribute and values_excluded, but the filters where not applied. This has been corrected so it is back to the spirit of the intended functionality. An adjustment to derive_level_paths was made to match the fix. 

Further Note: A review of other method in this class suggest use of the those filters have not been applied in most methods and will likely need to be corrected at some point. While those fixes are very quick and simple (matching the adjusted "from_file" method, significant testing would be required. This testing can be done by upgrading its unittest file.

src/gms/stream_branches.py
```diff
@@ -19,6 +19,20 @@
 
 class StreamNetwork(gpd.GeoDataFrame):
 
+    '''
```
BradfordBates-NOAA, line 22: @RobHanna-NOAA

src/gms/stream_branches.py
```diff
@@ -58,16 +72,14 @@ def from_file(cls, filename, branch_id_attribute=None,values_excluded=None,
         if verbose: 
             print('Loading file')
             
-        dataframe = gpd.read_file(filename,*args,**kwargs)
+        raw_df = gpd.read_file(filename,*args,**kwargs)
         filtered_df = gpd.GeoDataFrame()
-
```
RobHanna-NOAA: Pre-commit, this method accepted the branch_id_attribute and values_excluded params, but they were not actually be used to filter data. This has now been corrected but for this method only. Other methods in this class have this same issue and will someday need to be fixed.

BradfordBates-NOAA: Thanks for catching and explaining this, Rob.

BradfordBates-NOAA:
@RobHanna-NOAA Thanks for the thorough description. Can you please use the tick marks, i.e `, to wrap file names and function, class, or method names? It helps with readability in the description.

RobHanna-NOAA:
Changelog.md updated and formatting for descriptions are now added.

Note: Branch can be deleted when merge complete.

CarsonPruitt-NOAA (approved):
Ran a 2-huc test on `gms_run_unit.sh` and a follow up test on `gms_run_branch.sh` to make sure that bugs weren't introduced in either step. Checked all outputs in QGIS and everything looks great.

```
 gms_run_unit.sh -u "02020005 12090205" -n carson_order_test -c /foss_fim/config/params_template.env -d /foss_fim/config/deny_gms_unit_default.lst -s -j 2
 gms_run_branch.sh -n carson_order_test -c /foss_fim/config/params_template.env -d /foss_fim/config/deny_gms_branches_min.lst -j 5
```

BradfordBates-NOAA:
Merged. Nice work @RobHanna-NOAA!

PR #549, opened 2022-03-02
Split flows fix

hohe12ly:
the new algorithm uses GeoPandas overlay with identity method to create splits. then these splits are ordered by lake IDs. a subsequence that goes in and out of a lake multiple times is modified to have at most three parts: the one going into the late, the one in the lake, and the one out of the lake. the new algorithm returns such splits as a GeoDataFrame for the next step 1.5km length-based splitting. That length-based splitting is unchanged.

Also, added one step after HUC8 overlay to remove empty geometries.

Changing the request to branch dev-FIM3


RobHanna-NOAA:
This PR has been dropped, it was part of FIM3 and is no longer applicable

Issue #567, opened 2022-03-17
[3pt] Bug in Aggregation Step of fim_run.sh


nickchadwick-noaa:
A recent FIM run ended in an Error which I believe is the aggregation step at the end of `fim_run.sh`
![image](https://user-images.githubusercontent.com/66972397/158854845-7a95e161-e05f-4a6c-80d0-407d63cae0c8.png)

The full log can be found at `previous_fim/fim_3_0_26_1_ms_c/logs/docker.log`

PR #570, opened 2022-03-23
[1pt] PR: Hotfix to address catchment aggregation with "src_calibrated" field

RyanSpies-NOAA:
Hotfix to address error with mismatched HUC8 catchment attributes when aggregating/appending. Added a check to create "src_calibrated" field if it doesn't exist for a HUC prior to appending. Address #567 

## Changes

- `src/aggregate_fim_outputs.py`: Added check to create "src_calibrated" field with "null" entries if the field doesn't exist for a HUC8. This address an error when appending catchment layers that have the field (HUC8 with calibration performed) with HUCs that do not have the field.

## Testing

1. Confirmed fix on HUC6 170501 (prior error in MS BED run). Usage: `fim_run.sh -u "17050101 17050102 17050103 17050104" -e MS -c foss_fim/config/params_template.env -n ryan_temp/bug_fix_catch_attrib -j 4 -v`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,8 +1,6 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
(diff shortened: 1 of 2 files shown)


BradfordBates-NOAA (commented):
Confirmed that the outputs were produced as expected.

PR #581, opened 2022-04-13
[1 pt] PR: Tool to analyze correlations in SRC error and various variables

CalebOliven-NOAA:
This pull request is to add the current version of the `correlation_analysis.py` tool to the fim tools library.
The analysis tool takes the output from the `collate_catchment_attributes.py` tool and performs two main analyses. The first is to generate correlations between all the variable present in the collate output. These include land cover, slope, etc. The second is to bin the rows of the collate tool output by huc4, and perform a correlation analysis on each one of those bins, then compile those results into an easy to read histogram. At present, the huc4 analysis portion can only operate on a single variable at a time. In the future, a nice addition would be to enable the script to perform multiple variables at a time. 

To run this code, use the following command within the correct tools folder within a docker container:

`python correlation_analysis.py -p "/data/temp/caleb" -in "/data/temp/caleb/master_data/ms_all_gauges_nlcd.csv" -var "lulc_2" -sep "yes"`

The flag `-sep` expects a yes or no string. If yes, it will take the input and rebin the rows into huc4's. If no, it will not.
Using the parent directory `data/temp/caleb` means that the folders and huc4's already exist. So running it on that parent directory with the `-sep` flag would be redundant. If running on a different parent directory, change the `-sep` flag to yes for the first time it is run. 

If the `-sep` flag is yes and the folders already exist, it will not break the code. It is just going to take a bit more time to recreate the huc4 csv's.
Run once with `-sep` flag set to yes before running with `-sep` flag set to no. Set to no and change variable to obtain histograms for the variable/error correlation  binned into 0.1 width bins.

-

## Testing

1.To test the code use command:
`python correlation_analysis.py -p "/data/temp/caleb" -in "/data/temp/caleb/master_data/ms_all_gauges_nlcd.csv" -var "lulc_2" -sep "yes"`


BradfordBates-NOAA:
@CalebOliven-NOAA It looks like the usage provided creates an empty plot, am I doing something wrong? Here is my usage:
`python correlation_analysis.py -p "/data/temp/caleb" -in "/data/temp/caleb/master_data/ms_all_gauges_nlcd.csv" -var "lulc_2" -sep "no"` 

lulc_2

BradfordBates-NOAA (changes requested):

BradfordBates-NOAA (approved):

Issue #586, opened 2022-04-21
[5pts] Create tools and code adjustments to assist in alpha testing analysis of GMS outputs.


RobHanna-NOAA:
During large scale testing of GMS for alpha testing (`synthesize_test_cases.py`) was attempted, some challenges were discovered.

When the full GMS BED was run a few months back, due to output sizes and processing times,  ~2,180 HUCs were processed in sets of 100 across two servers.  By default, `synthesize_test_cases.py` is looking for a centralized single output folder to do alpha testing.  Further, they needed to be processed with a valid `test_cases` folders in a specific relational path. Due to overall total size of GMS outputs, which was ~ 3.9 TB, it was very difficult, and some times impossible, to put them into a central location. Quick pre-tests also strongly suggested that processing of all ~2,180 GMS folders resulted in very slow alpha testing processing. 

Tools and processes need to be created that can extract only HUCs that have valid test data then alpha test process just related HUC (unit and branch) folders that are applicable. Creating a list of alpha test applicable HUCs is needed, then use that list to extract correct GMS processed HUC folders. `synthesize_test_cases.py` and its child python files rely on a key file in a key location called `gms_inputs.csv` to exist and match folders in it's directory. Without a valid `gms_inputs.csv` which matches only applicable alpha tested common folder HUCs (units and branches), processing was slow with thousands of errors. This is due the total of ~ 2,180 total HUC files and hundreds of thousands of branch files. Only 216 HUCs are valid for alpha testing. 

All January GMS folders are on production servers and not in locations where most staff can access them. Large volumes and processing times result in it being best to process in production servers instead of dev machines which were often previously used. 

Tools, enhanced command line outputs / tracking and new processes are required to fully process all alpha testing applicable valid GMS folders / files.

PR #587, opened 2022-04-21
[1pt] PR: New tools and adjustments for bulk gms processing for alpha tests

RobHanna-NOAA:
During large scale testing of GMS for `alpha testing` (synthesize_test_cases.py), some challenges were discovered. Tools and upgraded log / output tracking were required for `alpha testing` the large GMS processed folders. This was complicated by the fact that GMS output folders are likely not all in one directory or even one machine.

When alpha testing is completed, it adds a number of folders/files to a common `test_case` folder. The merging of folders and files from different FIM3 or FIM4/GMSprocessed folders is very tricky to amalgamate correctly and high risk. 

Extracting only GMS HUC (unit and branch) folders that were valid for alpha testing had a major impact on overall 
performance of `synthesize_test_cases.py`.  In order to figure out which output folders were required, one needed to create a list of HUCs that had valid test case data. The new `find_test_case_folder.py` file scans through the `test_cases` folder and creates a list of HUCs that have some test data. You can define the path and file name of the HUC list, ie) /data/inputs/huc_lists/huc_list_for_alpha_tests_22020420.lst. 

Now that we have a list of HUCs, we can run another new tool named `copy_test_case_folder.py` which uses the new HUC list to scan an output folder and copy out the HUC (unit and branches) folders into a common folder. During full BED testing of GMS, we had a total of 22 folders, each with appx 100 HUCs (units and branches). Running the `copy_test_case_folder.py` against all 22 folder (across 2 servers), allows only HUC folders that were applicable to alpha testing into a common directory. Note: It is not required to have all in one folder but make the total speed and tracking much better.

Now that we had the GMS output folders for HUCS that were `alpha testing` applicable, we needed to create a new _gms_inputs.csv_ file as this is required for running alpha testing.  Adding a command line param to the `copy_test_case_folder.py` would calculate a new _gms_input.csv_ for only HUCS in the amalgamated HUC alpha test folder. To get the new _gms_input.csv_, we needed to change the `aggregate_branch_lists.py` file to allow for more input parameters and less hardcoding. As the interface for `aggregate_branch_lists.py` changed, `gms_run_unit.sh` has to change as well.

Some additional logging was added in a couple of key places to help track progress including start dates/time, end dates/time and durations. This was added to `synthesize_test_cases.py`, `run_rest_case.py` as well as added back to `gms_run_unit.sh` and `gms_run_branch.sh`.

A few small adjustments were also made to `gms_run_unit.sh` and `gms_run_branch.sh` which could throw errors in certain overwrite / rerun conditions.

See the original [Issue 586](https://github.com/NOAA-OWP/inundation-mapping/issues/586). 

## Additions

- `tools/find_test_case_folders.py`
   - The creates a list of HUCs that have valid test data available for alpha testing. It scans through a stated directory as an argument (likely `test_cases` ) looking for all _first level_ folders following the convention of ** *_test_cases**. ie) usgs_test_cases.  Then scans through _each applicable child folder_ looking for a pattern of 8 numbers, underscore, then 3 to 7 characters. ie) 01020304_usgs.  The HUC value, ie) 01020304, is extract and added to a unique HUC value list, line delimited.
   
- `tools/copy_test_case_folders.py`
   - This tool allows for copying data from one folder to another and can use a HUC list to do it. Generally, it is used to extract only folders that are applicable to alpha testing to be consolidated to a central location for processing speed reasons. Using any .text or .lst list using values with line delimiters, it can scan through a root directory, submitted as an argument, looking for matching folder. ie) 01020304 folder in /outputs/jan_gms_run.  Note. While it will usually be a HUC list against a FIM3 or FIM4/GMS output folder, it is not mandatory. It simply looks for a folder name match at the first level under the root defined folder.  As that folder is found, it is copied including sub-files/folders, to another folder defined as a parameter. ie) /outputs/jan_gms_run_test_hucs.  Further, it has an optional argument that can generated a _gms_input.csv_ file in the target folder as the _gms_input.csv_ is required for `alpha testing`.

## Changes

- `src/gms/aggregate_branch_lists.py`
   - Was updated to allow for two new arguments which were previously hardcoded. Previously this file was used only by `gms_run_unit.sh` to create the _gms_inputs.csv_ file required for further processing steps. With adding the two new params, this tool can be re-used by tools as part of `alpha testing` via the new `copy_test_case_folders.py` to create a new _gms_inputs.csv_ file after coping folders to new locations.  

- `tools/synthesize_test_cases.py`
   - Was updated primarily for enhanced logging. It now includes outputs that show start date/time, end date/time and duration. During large scale use of the tool, progress tracking was limited.

- `tools/run_test_case.py`
   - An additional print statement at a key location helped with progress tracking and now shows what test_id (HUC and test type) is being processed.

- `gms_run_unit.sh`
   - This was upgraded for two things: 
      - Change it call to aggregate_branch_lists.py to add in the two new additional input arguments.
      - Upgrades for output to help with overall start date/time, end date/time and duration of the entire gms_run_unit.sh command. This helps track overall performance.

- `gms_run_branch.sh`
   - Upgrades for output to help with overall start date/time, end date/time and duration of the entire gms_run_unit.sh command.  This helps track overall performance.

## Testing
A couple of different test flows were done and most individual steps were tested a number of times independently to validate outputs, different input params, error conditions, etc.
One test scenario was:
    - This was tested against processing of two HUCs, one at time, through `gms_run_unit.sh` and `gms_run_branch.sh` to validate the upgraded output, but also that the changes related to creating the _gms_inputs.csv_ was done correctly. By running the two HUCs separately, two separate output folders were created.
   - Running the `find_test_case_folders.py`  tool created a master list of HUCs that had test data. This file was compared to a manually created list by another staff member to ensure it was correct. Running the tool helped prevent a manual processing of discovering valid HUCs.
   - Using the new valid test case HUC list and using the `copy_test_case_folders.py`, it was run twice against the two new test HUC folders. When running the second folder, the input param for creating the manditory `gms_inputs.csv` was included and validated by hand.
   - `synthensize_test_case.py` was run against the new combined folder with the outputs into the `test_cases` folders being evaluated at a high level, aka.. were expected folders added in correct places in the `test_cases` folder. 
   - The overall metrics output file from `synthensize_test_case.py` was not independently evaluated. This was due to the core logic of `synthensize_test_case.py` not changing. Advanced validation of the metrics file is being evaluated separately and is expected to be fine.

The entire process and tools in order to create alpha test data was rrun twice. In past months, a full BED GMS test was done which created 22 folders across two servers.  For each server, and its appx 11 folders, the entire process was run in order to create full alpha test results.  Validation at each step was done during the full alpha test run for each server. Once each server was completed, the final results were copied back to merged central server `test_cases` folder.  Merging back to a centralized `test_cases` folder was still tricky but significantly easier than validation of 22 folders from the original GMS output folders. 

Based on the initial list of 216 HUCs that appeared on the huc list for alpha testing, it was extremely difficult to validate if the correct number of total new `test_cases` folders were created. However, spot checking of a number of HUCs from both of the original two servers, suggest the outputs are fine.  No unexpected or unaccounted errors were found during final usage of the tools and processes.

### DIFF
--- a/tools/run_test_case.py
+++ b/tools/run_test_case.py
@@ -36,6 +36,8 @@ def run_alpha_test( fim_run_dir, version, test_id, magnitude,
     # make bool
     calibrated = bool( calibrated )
 
+    print(f"run_alpha_test: test_id = {test_id}")
+
     if (model == "MS") & (fr_run_dir is None):
         raise ValueError("fr_run_dir argument needs to be specified with MS model")
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,31 @@ All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 
+## v4.0.4.1 - 2022-05-02 - [PR #587](https://github.com/NOAA-OWP/inundation-mapping/pull/587)
+
+While testing GMS against evaluation and inundation data, we discovered some challenges for running alpha testing at full scale. Part of it was related to the very large output volume for GMS which resulted in outputs being created on multiple servers and folders. Considering the GMS volume and processing, a tool was required to extract out the ~215 HUC's that we have evaluation data for. Next, we needed isolate valid HUC output folders from original 2,188 HUC's and its 100's of thousands of branches. The first new tool allows us to point to the `test_case` data folder and create a list of all HUC's that we have validation for.
+
+Now that we have a list of relavent HUC's, we need to consolidate output folders from the previously processed full CONUS+ output data. The new `copy_test_case_folders.py` tool extracts relavent HUC (gms unit) folders, based on the list created above, into a consolidated folder. The two tools combine result in significantly reduced overall processing time for running alpha tests at scale.
+
+`gms_run_unit.sh` and `aggregated_branch_lists.py` were adjusted to make a previously hardcoded file path and file name to be run-time parameters. By adding the two new arguments, the file could be used against the new `copy_test_case_folders.py`. `copy_test_case_folders.py` and `gms_run_unit.sh` can now call `aggregated_branch_lists.py` to create a key input file called `gms_inputs.csv` which is a key file required for alpha testing.
+
+A few other small adjustments were made for readability and traceability as well as a few small fixes discovered when running at scale.
+
+## Additions
+
+- `tools/find_test_case_folders.py`: A new tool for creating a list of HUC's that we have test/evaluation data for.
+- `tools/copy_test_case_folders.py`: A new tool for using the list created above, to scan through other fully processed output folders and extract only the HUC's (gms units) and it's branches into a consolidated folder, ready for alpha test processing (or other needs).
+
+## Changes
+
+- `src/gms/aggregate_branch_lists.py`: Adjusted to allow two previously hardcoded values to now be incoming arguments. Now this file can be used by both `gms_run_unit.sh` and `copy_test_case_folders.py`.
+- `tools/synthesize_test_cases.py`: Adjustments for readability and progress status. The embedded progress bars are not working and will be addressed later.
+- `tools/run_test_case.py`: A print statement was added to help with processing progess was added.
+- `gms_run_unit.sh`: This was adjusted to match the new input parameters for `aggregate_branch_lists.py` as well as additions for progress status. It now will show the entire progress period start datetime, end datetime and duration. 
+- `gms_run_branch.sh`: Also was upgraded to show the entire progress period start datetime, end datetime and duration.
+
+<br/><br/>
+
 ## v4.0.4.0 - 2022-04-12 - [PR #557](https://github.com/NOAA-OWP/inundation-mapping/pull/557)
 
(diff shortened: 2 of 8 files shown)


BradfordBates-NOAA:
@RobHanna-NOAA It looks like the PR title is cut off with a "..."?

RobHanna-NOAA:
Does this work (title)?

tools/run_test_case.py
```diff
@@ -36,6 +36,8 @@ def run_alpha_test( fim_run_dir, version, test_id, magnitude,
     # make bool
     calibrated = bool( calibrated )
 
+    print(f"run_alpha_test: test_id = {test_id}")
```
CarsonPruitt-NOAA, line 39: This might mess with the progress bar that's set up in `synthesize_test_cases.py`

CarsonPruitt-NOAA (approved):
The new additions look good. I really like hoe you added the ability to use `aggregate_branch_lists.py` to recreate the _gms_inputs.csv_. I would have used that in the past when I needed a new _gms_inputs.csv_.

CarsonPruitt-NOAA (approved):

PR #597, opened 2022-05-16
[1pt] PR: Apply BARC for select FR streams

RyanSpies-NOAA:
This PR enables BARC to be implemented to FR flow lines to better match the MS inundation where FR & MS overlap. This PR uses an input env parameter to control which stream orders BARC is applied to for both FR & MS (currently set to SO 4-9). Address #585 

## Additions

- `tools/copy_huc_dirs.py`: New utility script to copy a list of HUC fim output files from src directory to dst directory by feeding the script a list of HUC ids. This streamlines the development process when testing code modifications and performing the subsequent FIM evaluations.

## Changes

- `config/params_template.env`: Changed `ignore_streamorders` parameter to use a list of stream orders that will be ignored within BARC routine
- `src/bathy_src_adjust_topwidth.py`: Revised to read a list of stream orders and then mask the BARC calculations for select stream orders
- `src/utils/shared_variables.py`: Included a copy of the BARC input parameters to allow user to run BARC as a stand alone script (outside `fim_run.sh`) 

## Testing

1. Performed a multitude of BARC stream order filtering iterations to identify the ideal FIM performance. Currently, the default solution will impliment BARC for both MS & FR resolutions for stream orders 4-9.
RAS2FIM alpha evaluation:
image
BLE alpha evaluation:
image

2. Performed qualitative evaluation of evaluation inundation outputs to confirm that FR inundation more closely agrees with the MS inundation in overlap areas (see examples below).

## Screenshots
Problem: With the current FIM3 version FR FIM extents typically produces larger inundated area compared to the MS inundation for the same flow line. BARC is effectively reducing the inundation extent for MS flow lines. With this configuration the FR + MS composite effectively negates the BARC influence for MS flows lines. This is highlighted by the distinct FR vs. MS inundation outputs in the image below (using same flow for both FR and MS).
image

Solution: Implementing BARC for both FR and MS results in similar synthetic rating curves and resulting inundation extents where FR and MS flow lines overlap. This ensures consistency between the two resolutions. The bankfull regression performance is notably less reliable for smaller streams (stream orders <4); therefore, the new approach limits BARC implementation to stream orders 4, 5, 6, 7, 8, and 9.
image
image

Mainstem stream order histogram:
image

Full resolution stream order histogram:
image

Noted several locations where gaps in the MS inundation (bad crosswalk?) are filled with the FR inundation. Suggest looking into the root cause of the MS gaps.
image

## Todos

- [ ] Currently running a FR BED simulation with the new code
- [ ] Updated changelog

### DIFF
--- a/config/params_template.env
+++ b/config/params_template.env
@@ -32,7 +32,7 @@ export thalweg_stg_search_max_limit=3 # Threshold: Stage value limit below which
 export bathy_xs_area_chg_flag=1 # Flag: Cross section area limit to cap the amount of bathy XS area added to the SRC. Limits the bathy_calc_xs_area/ BANKFULL_XSEC_AREA to the specified threshold
 export bankful_xs_area_ratio_flag=10 # Flag: Identify bogus BARC adjusted values where the regression bankfull XS Area/SRC bankfull area is > threshold (topwidth crosswalk issues or bad bankfull regression data points??)
 export thalweg_hyd_radius_flag=10 # Flag: Idenitify possible erroneous BARC-adjusted hydraulic radius values. BARC discharge values greater than the specified threshold and within the thal_stg_limit are set to 0
-export ignore_streamorders=10 # Ignore BARC calculations for streamorders >= this value (10 is Mississippi R)
+export ignore_streamorders=[1,2,10] # Ignore BARC calculations for streamorders >= this value (10 is Mississippi R)
 
 #### estimating bankfull stage in SRCs ####
 export src_bankfull_toggle="True" # Toggle to run identify_bankfull routine (True=on; False=off)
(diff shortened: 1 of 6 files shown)


RyanSpies-NOAA:
@CarsonPruitt-NOAA I confirmed full domain (BED) simulations for both FR and MS completed successfully (**all SRC post-processing included - BARC, VMann, point calibration**). The alpha evaluation metrics for both MS & FR show minimal or no differences to the existing FIM 3.0.28.1 metrics. See examples below:

### Mainstem evaluation comparison:
csi_ras2fim_ms
csi_nws_ms

### Full-Res evaluation comparison:
csi_ras2fim_fr
csi_ifc_fr
csi_usgs_fr

CarsonPruitt-NOAA (dismissed):

CarsonPruitt-NOAA (approved):

Issue #519, opened 2022-01-31
[5pt] Run_by_unit does take multiple huc comma delimited


RobHanna-NOAA:
This functionality existed in FIM3 in its version (as part of fim_run.sh)

Error message:
- fails on multiple commented hucs
Traceback (most recent call last):
File "/foss_fim/src/check_huc_inputs.py", line 61, in
check_hucs(**args)
File "/foss_fim/src/check_huc_inputs.py", line 48, in check_hucs
__check_for_membership(hucs,accepted_hucs)
File "/foss_fim/src/check_huc_inputs.py", line 41, in __check_for_membership
raise KeyError("HUC {} not found in available inputs. Edit HUC inputs or acquire datasets and try again".format(huc))
KeyError: 'HUC 12030107,06030002,03070105,05010009 not found in available inputs. Edit HUC inputs or acquire datasets and try again'

Note: Documentation in the run_by_unit.sh suggest it should work.

Note: This might have not worked in FIM3

RobHanna-NOAA:
In testing, it was discovered that a comma delimited is not supposed to work. However it does work as space delimited with a set of double quotes around the whole thing.  ie ) -u "07080101 05030204 13030202" (will work).  All that is required is to upgrade the "help" notes for run_by_unit.sh.

PR #612, opened 2022-06-07
[1pt] PR: Alpha Test Refactor

CarsonPruitt-NOAA:
A few things were addressed in this PR:
- [Issue #602](https://github.com/NOAA-OWP/inundation-mapping/issues/602) : Applying the `test_case` class structure from the `dev-fim3` branch along with extra inline commenting, formatting, upgrading inline and verbosity print statements. This changes increases readability, traceability and performance of doing alpha tests.
- [Issue 519](https://github.com/NOAA-OWP/inundation-mapping/issues/519): Run_by_unit does take multiple HUC comma delimited. Stumbled across this issue during testing and fix it as part of this PR.
- Added a cleaner system for adding print lines when the verbose flag is being used.

## Changes

- `gms_run_branch.sh`:  Updated help instructions to about using multiple HUCs as command arguments.
- `gms_run_unit.sh`:  Updated help instructions to about using multiple HUCs as command arguments.
- `src/utils`
    - `shared_functions.py`: 
       - Added a new function called `vprint` which creates a simpler way (and better readability) for other python files when wanting to include a print line when the verbose flag is on.
       - Added a new class named `FIM_Helpers` as a wrapper for the new `vprint` method. For it to be used in other code files, the command now becomes ie) fh.vprint("Creating output files", verbose), instead of an earlier convention saying "if verbose then: print... ). Note. Some files already had their own version of this helper type of code under functions named similar to `def __print(message, verbose.....etc`. This new classes removes the duplication, standardizes it and make is available to all python files.
       - With the new `FIM_Helpers` class, a previously existing method named `append_id_to_file_name` was moved into this class making it easier and quicker for usage in other classes.
       
- `tools`
    - `composite_inundation.py`: Updated its usage of the `append_id_to_file_name` function to now call the`FIM_Helpers` method version of it.
    - `gms_tools`
       - `inundate_gms.py`: Updated for its adjusted usage of the `append_id_to_file_name` method, also removed its own `def __vprint` function in favour of the `FIM_Helpers.vprint` method. 
       - `mosaic_inundation.py`: 
          - Added adjustments for use of `append_id_to_file_name` and adjustments for `fh.vprint`.
          - Fixed a bug for the variable `ag_mosaic_output` which was not pre-declared and would fail as using an undefined variable in certain conditions.
    - `run_test_case.py`: Ported `test_case` class from FIM 3 and tweaked slightly to allow for GMS FIM. Also added more prints against the new fh.vprint method. Also added a default print line for progress / traceability for all alpha test regardless if the verbose flag is set.
    - `synthesize_test_cases.py`: Ported `test_case` class from FIM 3.
- `unit_tests`
   - `shared_functions_unittests.py`: Update to match moving the `append_id_to_file_name` into the `FIM_Helpers` class. Also removed all "header print lines" for each unit test method (for output readability).
   
## Testing

1.  A number of single tests using one and multiple HUCs were performed.  This was done in sets of with the using all stream orders as well as filtering stream orders 1 and 2.
2. A full `test domain` run was done using all HUCs that we have benchmark data for. Then a comparision of the outputs and metrics was compared to an earlier test domain run prior to this PR. Both sets were identical, validating that the changes in this PR have not affected core functionality.


RobHanna-NOAA:
I updated a few things (as I mentioned)
1) I added a new class and method to make verbose printing cleaner for inline code in the src/utils/shared_functions.py.
2) I updated some of the files to add extra verbose printing as/when needed. When testing in the UCS machines, it was harder to see what was failing.
3) A bug existed in the code in mosaic_inundataion.py for the variable `ag_mosaic_output` not being defined in certain conditions. Further, the exception was not being trapped or logged correctly. 
4) It also exposed a problem of trying to abort the entire command via Ctrl-C (Keyboard interrupt). Ctrl-C would stop on the one single process (of a multi process set). Hence if you have 10 multi procs running, you would have to do Ctrl-C 10 times. Otherwise, after 10-60 seconds another proc  would continue. One or two Ctrl-C's will now kill everything immediately.
5) I cleaned up a few tidbits for styling.
6) Issue 519 (Run_by_unit does take multiple huc comma delimited) existed and I stumbled across of it. Turns out it was just unclear inline documentation (help command). I fixed this and this issue can be closed.

Can you double check my changes for me?

I am doing a full Test Domain run on UCS2 to ensure all is well which we can do before fully approving and merging the pull request.  

I will update the pull request notes at the top.

I can update the Changelog.md when we get there.

Many thanks.

RobHanna-NOAA (dismissed):

PR #615, opened 2022-06-10
[1pt] PR: Alpha metrics by catchments (FIM3)

CalebOliven-NOAA:
This PR adds the functions and changes necessary to aggregate alpha metrics by catchment for the FIM3 product. The function `test_case_by_hydroid.py` performs this aggregation. It does so utilizing the ogr library.

Additional functions are involved in the process and are also uploaded to the `tools` folder.
These include: 

- `pixel_counter.py`
This script isolates the number of pixels per class of a raster within the outlines of
one or more polygons and displays them in a table. It accomplishes this by rasterizing the vector file,
masking out the desired areas of both rasters, and then summarizing them in a dataframe. It makes use
of the gdal, numpy, and pandas function libraries.
- `pixel_counter_wrapper.py`
This function sets up processing of the process_zonal_stats() function.
- `pixel_counter_functions.py`
Contains functions for processing raster data, specifically the Land Use Land Cover dataset. It also contains a function `get_mask_value_counts` which is the main function used for this specific process of digesting Alpha metrics. This function processes the `agreement_raster` produced by the alpha test. 

`test_case_by_hydroid.py` outputs both a CSV file and a Geopackage containing Alpha metrics. The `-comp` composite flag dictates whether the composite metrics are applied. If applied, the comp flag will apply the composite Alpha Metrics to the boundaries defined by the version input geopackages.

The `-v` version flag dictates which geopackages will be used as inputs, as well as which validation metrics will be applied in absence of the `-comp` flag.

Example for running MS configuration: `python3 test_case_by_hydroid.py -b"all" -v"fim_3_0_24_14_ms" -c"/path/to/output.gpkg"`

Example for running COMP configuration against FR catchment boundaries: `python3 test_case_by_hydro_id.py -b"all" -v"fim_3_0_36_0_fr_c" -c"/path/to/output.gpkg" -comp`

When running the tool, give it as an output path, a `/data/path/file.gpkg` with the gpkg suffix intact. Current method is to use sting operations to get the name for the csv output from the gpkg output string, so please input a `.gpkg` suffix on the output string.

Here is an image of an pre-processing agreement raster:
image

The function then takes a gpkg for catchment boundries and counts pixels for each hydroid. 
Below is an image of a geopackage displaying CSI for each catchment:
image

## Additions

-`tools/pixel_counters.py`   This tool is used for a function `perform_zonal_stats(huc_gpkg,agree_rast)` Changes to this function were added to facilitate its intended use within `test_case_by_hydroid.py` Specifically `agree_rast` was added as a input to the function. This agreement raster is found within fim data and defines, pixel by pixel, inundation accuracy. 
The input huc_gpkg is a single huc8 geopackage, the second input argument must be input as a dict.
For the purposes of assembling the alpha metrics by hydroid, always use agreement_raster total area agreement tiff.
This function is called automatically. Returns stats dict of pixel counts.

-`tools/pixel_counter_functions.py` Added a function called `get_mask_value_counts` that counts the pixels from inputs that have different mask values. This gives true/false and positive/negative values for pixels on a catchment scale.

-`tools/pixel_counter_wrapper.py` This is a necessary function to make the other added pixel counter functions work. 

-`tools/test_case_by_hydroid.py` This is the main function for which this pull request is being generated.

## Output

-image

## Testing

1. To run the function with faster speed, use 'usgs' as the benchmark. Otherwise use the desired param. All will use all available data per the version specified. 
2. Use command `python3 test_case_by_hydro_id.py -b"all" -v"fim_3_0_36_0_fr_c" -c"/data.gpkg" -comp`


BradfordBates-NOAA:
@CalebOliven-NOAA It looks like there is some funky merge stuff going on in your branch. I noticed at least once instance in the Changelog.

image

BradfordBates-NOAA:
@CalebOliven-NOAA Recommend clarifying the `-c` argument. Right now it says `Path to folder to hold exported csv file.` But the code appears to use it as a CSV path, not a directory path.

CalebOliven-NOAA:
Good comment. I have changed the language. If it is still confusing let me know, and I will do a bigger rewrite.

BradfordBates-NOAA:
@CalebOliven-NOAA It looks like the Changelog is still not right. There could be other files affected by the merge as well.

RobHanna-NOAA:
Carson and I can fix the changelogs. They have been bumpy on each run.

CarsonPruitt-NOAA (approved):
Reviewed and tested by @BradfordBates-NOAA

BradfordBates-NOAA:
I approve.

Issue #609, opened 2022-06-01
[2pt] API updates to enable PostgreSQL database connection


RyanSpies-NOAA:
To-do list to allow production machine to run fim_run.sh with new postgres database for SRC calibration:
- [ ] Add function to copy the params.env file when pulling the repo
- [ ] Rebuild docker image with psycopg2

RyanSpies-NOAA:
@nickchadwick-noaa would it be possible to wrap up these modifications before the end of the day? It would be great to run a BED test simulation on the production machine over the weekend.

PR #618, opened 2022-06-24
[1pt] PR: PostgreSQL DB solution for calibration point data 

RyanSpies-NOAA:
These changes introduce a PostgreSQL database solution for storing, processing, and accessing the point-based calibration data previously stored in a gpkg. The PostgreSQL enables faster processing and a flexible solution for continuously increasing the number of calibration points. Addresses #572 & #609 

## Additions

- `tools/calibration/README.md`: documentation and tips for working with the PostgreSQL DB
- `tools/calibration/start_db.sh`: bash shell script for creating the docker container for the database (only needs to be executed if the container is closed or disconnected - see instructions in README)

## Changes

- `config/params_template.env`: added a variable with the filepath location for the sensitive variables
- `fim_run.sh`: added two ogr2ogr commands to populate the PostgreSQL database with the point and huc boundary spatial data
- `src/src_adjust_spatial_obs.py`: revised to ingest the PostgreSQL database, perform a spatial query to create a list of all HUCs containing calb points, query points to each huc, and pass a dataframe to the `src_roughness_optimization.py` workflow
- `src/src_adjust_usgs_rating.py`: simplified the `-debug` flag (action='store_true')
- `src/src_roughness_optimization.py`: added placeholder variable creation (this is only needed to run the script for older FIM versions as they are now created by default in `add_crosswalk.py`)
- `api/node/updater/updater.py`: added code that copies the `params.env` file to the API job's working directory for database connection purposes.

## Testing

1. Successfully completed MS & FR test runs for all evaluation HUCs and verified no/negligible changes to the alpha evaluation metrics (see comparison plots below)
2. Run time improvement: 
image

## Screenshots
MS alpha metrics comparison (USGS benchmark sites):
csi_aggr_usgs_ms

FR alpha metrics comparison (NWS/AHPS benchmark sites): 
csi_aggr_nws_fr

## Notes

- NOTE: this new configuration requires user to have a new file `config/params.env` to store sensitive information for accessing the PostgreSQL database docker container. **The `params.env` is not tracked in the repo so all users will need to create/copy this file to their local repo location prior to running.**
- The only noticeable change to the SRC/FIM performance is the result of a slight modification to how the points are "clipped" to a given HUC. Previously we clipped the .gpkg points to the buffered HUC domain whereas the PostgreSQL DB spatial query limits the points to those located directly within the HUC boundary (non-buffered HUC domain). This doesn't appear to have a significant impact the resulting FIM performance, but we could look to modify this in the future if needed.

## Todos

- update changelog prior to merge


RyanSpies-NOAA:
Updated changelog for merge

CarsonPruitt-NOAA (approved):

Issue #603, opened 2022-05-23
[21pt] Develop optional CatFIM technique to use direct stage-to-HAND method


CarsonPruitt-NOAA:
For each CatFIM flow, convert the AHPS stage value to Water Surface Elevation (WSE), then to FIM stage. Inundate the REM at the calculated FIM stage to produce the CatFIM data. This approach will remove any error that is produced by using the FIM synthetic rating curves.

BradfordBates-NOAA:
@CarsonPruitt-NOAA Just to confirm--this issue is to rework the CatFIM technique to the "Crane Method" for FIM3 right?

CarsonPruitt-NOAA:
Yes, that's correct. We will port this to FIM v4 later on when it's the version being used in HydroVIS.

BradfordBates-NOAA:
@CarsonPruitt-NOAA Can you remind me what the datum accuracy threshold is during the `usgs_elev_table.csv` production?

CarsonPruitt-NOAA:
Yes, the threshold is 1 foot. Anything with accuracy greater than that will be left out.

PR #619, opened 2022-06-24
[1pt] PR: CatFIM B Option (Water Surface Elevation Method)

BradfordBates-NOAA:
This PR adds a new option to the `/tools/generate_categorical_fim.py` script `-a` that allows for an alternate method for producing inundation maps at the official AHPS flood categories.

This PR addresses #603.

## Summary

I used FIM 3.0.32.0 for this testing. 3.0.32.0 includes calibration at USGS/NWS sites using both the available synthetic rating curves and the spatial “Alpha” benchmark data.

### CatFIM A (Original Method)
CatFIM A refers to the original CatFIM methodology, i.e. plugging the NWS threshold flows directly into the HAND synthetic rating curves, generating a new stage value different than the official threshold stage, and mapping it to the HAND grid 5 miles up and downstream of sites.

### CatFIM B (Alternate Method)
CatFIM B refers to the water surface elevation technique (see next slide). It determines the stage associated with a NWS threshold, converts the stage to WSE, applies the datum conversion if necessary (to match the HAND DEM datum). Then the HAND SRCs are converted from stage/flow pairs to WSE/flow pairs. The WSE/flow pair is used to interpolate the flow necessary to yield the NWS WSE. The HAND flow is then run through inundation.py to yield an inundation map 5 miles up and downstream of sites. I interpolate a HAND flow value because the WSE derived for a single gage is only useful for its home HAND catchment; the WSEs are not directly usable in up- or downstream catchments. By determining the flow associated with the WSE and then mapping it, it becomes possible to model “the same” conditions necessary to yield the WSE in nearby catchments.

## Changes

- `generate_categorical_fim_flows.py`: Added options `-a` and `-f`. `-a` is interpreted as a bool and if provided by a user will run the CatFIM B method. If `-a` is passed, then the `-f` flag (fim version full path) is required so that necessary input layers can be accessed.
- `generate_categorical_fim.py`: Added `-a` option to run the alternate CatFIM method.

## Testing

1. Run `generate_categorical_fim.py`. For example ```python generate_categorical_fim.py -f /data/previous_fim/fim_3_0_32_0_ms_c -a -j 6```

### Workflow
image

image

See #603 for screenshots of results.

## Notes

- There are much fewer mapped sites in CatFIM B than CatFIM A because over 500 HUCs not having sites with acceptable amounts of datum accuracy. See Todos below for next steps. 

## Todos

- After conversation with Crane, we will need to modify `usgs_gage_crosswalk.py` to output `usgs_elev_table.csv` files for all HUCs, regardless of datum accuracy, and include datum and horizontal accuracy information in the table as additional attributes. After that, `generate_categorical_fim.py` will be able to produce maps for over 500 more HUC8s, and the accuracy information will be added to the attribute table so that users can see and query the library by datum accuracy.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,17 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v3.0.33.0 - 2022-06-30 - [PR #619](https://github.com/NOAA-OWP/cahaba/pull/619)
+
+This PR adds a new option to the `/tools/generate_categorical_fim.py` script `-a` that allows for an alternate method (CatFIM B) for producing inundation maps at the official AHPS flood categories. Similar to the original CatFIM method (CatFIM A), CatFIM B produces inundation maps at the official AHPS categories, but instead of creating maps from the official AHPS discharges, maps are produced using the official stage values, which are converted to water surface elevation.
+
+## Changes
+
+- `generate_categorical_fim_flows.py` : Added options `-a` and `-f`. `-a` is interpreted as a bool and if provided by a user will run the CatFIM B method. If `-a` is passed, then the `-f` flag (fim version full path) is required so that necessary input layers can be accessed.
+- `generate_categorical_fim.py`: Added `-a` option to run the alternate CatFIM method.
+
+<br/><br/>
+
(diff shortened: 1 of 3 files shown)


BradfordBates-NOAA:
@CarsonPruitt-NOAA Just pinging this. Are there any steps before merging?

Issue #617, opened 2022-06-17
[8pt] HUC 04010201 bug causing hydrotable file size issue


RyanSpies-NOAA:
Appears that past several FIM3 versions have an issue with creating a very large hydrotable (~200+ gb). The table is very large because there are thousands of duplicate rows in the SRCs. Underlying root cause appears to be related to input geometry/vectors as there are numerous dem derived reaches and catchments with the same hydroid value. 

![image](https://user-images.githubusercontent.com/69868854/174317114-fb36bfb5-f583-4137-8fd2-ac6bfd4694bf.png)

hydrotable with showing duplicate rows:
![image](https://user-images.githubusercontent.com/69868854/174317477-51dcb85d-3f66-40f5-a9e4-3fc6904bb497.png)

BradfordBates-NOAA:
Taking notes.

It appears that this bug began sometime after version 3.0.28.1 (hydroTable size = 777KB).

PR #628, opened 2022-07-11
[1pt] PR: Addresses duplicate DEM derived streams and hydroid bug

BradfordBates-NOAA:
This merge adds three lines to `add_crosswalk.py` to drop duplicates from pandas dataframes. In prior versions, the duplicate features were causing the `hydroTable.csv` to blow up in size. This addresses #617.

I suspect this bug was introduced by a recent change in pandas or geopandas versions, as I was able to reproduce the issue on previous versions (3.0.7.0) where the issue did not used to occur in versions past. More testing would need to occur to confirm this, though.

## Changes

- `add_crosswalk.py`: Added pandas `drop_duplicates` to the `output_hydro_table` and `output_flows` and `output_src` dataframes to remove duplicates. This seems to have addressed the bug. 

## Testing
I tested this feature branch at the BED scale using the UCS2 API and confirmed all nonzero exit codes.

Screenshot of outputs for 04010201. Notice that the `hydroTable.csv` is not 200GB.
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,7 +87,7 @@ ENV PATH="$projectDir:${PATH}"
 ## install python 3 modules ##
 COPY Pipfile .
 COPY Pipfile.lock .
-RUN pip3 install pipenv && PIP_NO_CACHE_DIR=off PIP_NO_BINARY=shapely,pygeos pipenv install --system --deploy --ignore-pipfile
+RUN pip3 install pipenv==2022.4.8 && PIP_NO_CACHE_DIR=off PIP_NO_BINARY=shapely,pygeos pipenv install --system --deploy --ignore-pipfile
 
 ## RUN UMASK TO CHANGE DEFAULT PERMISSIONS ##
 ADD ./src/entrypoint.sh /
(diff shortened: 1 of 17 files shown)


RyanSpies-NOAA:
After inspecting the hydrotable.csv for 04010201, I'm still seeing issues with duplicates. This is causing the `hydroTable.csv` file size to be significantly inflated (currently ~1gb and should be ~100mb). It appears the duplicates are introduced between the `src_base.csv` and the `src_full_crosswalked.csv`. I can see the duplicates (lesser impact) in the `src_full_crosswalked.csv` and then the issue is inflated with the `hydroTable.csv`. I suggest trying to drop the duplicates from the `src_full_crosswalked.csv` to see if that resolves the downstream impacts.

BradfordBates-NOAA:
Fixed the issue thanks to Ryan's suggestion to remove duplicates from the `src_base`

RyanSpies-NOAA (approved):
I was able to successfully reproduce the outputs for huc 04010201 and did not see any duplicates in the `src_full_crosswalked.csv` or `hydroTable.csv`. Approved for merge.

BradfordBates-NOAA:
@CarsonPruitt-NOAA This is ready to be merged.

Issue #630, opened 2022-07-12
[5pt] Get gms_pipeline working. It will process gms_run_unit and gms_run_branch in one kickoff call


RobHanna-NOAA:
The gms_pipline.sh file already exists and needs to be updated. Update parameters, add a few new ones, and remove some functionality (aka.. don't let it kick off some subsquent processing such as alpha test). 

Also add a feature where after the gms_run_unit portion has been processed, count the number of unit log error files and figure out the percent pass/failed. If it surpasses a given percent, then do not continue to gms_run_branch.

This will make it easier for development and testing use, but also can be used in future enhancements such as the API V2 tool.

PR #641, opened 2022-07-27
[1pt] Updates to simplify usage of fim4 via gms_pipeline.sh

RobHanna-NOAA:
A file called `gms_pipeline.sh` already existed but was unusable. This has been updated and now can be used as a "one-command" execution of the fim4/gms run. While you still can run `gms_run_unit.sh` and `gms_run_branch.sh` as you did before, you no longer need to.  

With `gms_pipeline.sh` executing processed of both units (HUCs), then branches, there was a need to have a new tool that could be run between the unit and branch processing.  The new tool is called `check_unit_errors.py` and starts by being told the number of units (HUCs) being initially processed.  If there are too many errors, it will block continuing on to the branch processing. As it can take over 18 hours to process and users will not monitor the entire process, this will stop the system much earlier so problems can be addressed.  The current rules (both configurable) for stopping branch processing are:
- Only stop if there is at least 10 unit fails. We don't want the system to fail after just one fail.
- Stop if there are 10% or more failures. 

`gms_pipeline.sh` can run with minimal input and defaults most parameters.  Only the runtime name and HUC(s)/HUC list are now required. Note: gms_pipeline defaults to assume that stream orders 1 and 2 are to be dropped.

`gms_run_unit.sh` and `gms_run_branch.sh` were also improved to simplify inputs so that they too now only need the runtime name and HUC(s) / HUC list. Note: For these two files, the default is to NOT drop stream orders 1 and 2 for backward compatibility reasons. 

This PR is related to [Issue 630](https://github.com/NOAA-OWP/inundation-mapping/issues/630). 

There are a number of minor adjustments as well, noted embedded below in file changes.

New FIM4/gms processing can be done with simply `gms_pipeline.sh -n <output name> -u <HUC(s) or HUC list path>`

## Additions

- `src`
    - `check_unit_errors.py`: as described above.
- `unit_tests`
    - `check_unit_errors_unittests.py` and `check_unit_errors_params.json`: to match new file.    

## Changes

- `README.md`:  Updated text for FIM4, gms_pipeline, S3 input updates, information about updating dependencies, misc link updates and misc text verbage.
- `gms_pipeline.sh`: as described above.
- `gms_run_unit.sh`: as described above. Also small updates to clean up folders and files in case of an overwrite.
- `gms_run_branch.sh`: as described above.
- `src`
     - `utils`
         - `fim_enums.py`:  FIM_system_exit_codes renamed to FIM_exit_codes.
         - `shared_variables.py`: added configurable values for minimum number and percentage of unit errors.
    - `bash_functions.env`:   Update to make the cumulative time screen outputs in mins/secs instead of just seconds.
    - `check_huc_inputs.py`:  Now returns the number of HUCs being processed, needed by `gms_pipeline.sh` (Note: to get the value back to a bash file, it has to send it back via a "print" line and not a "return" value.  Improved input validation, 
- `unit_tests`
   - `README.md`: Misc text and link updates.

## Removals

- `config\params_template_calibrated.env`: No longer needed. Has been removed already from dev-fim3 and confirmed that it is not needed.

## Testing

Over 40 tests were performed for a wide number of scenarios for all parameters being feed into `gms_pipeline.sh`, `gms_run_unit.sh` and `gms_run_branch.sh`. These included missing params, bad params and validating that defaults worked when expected.  A batch of 20 HUCs was tested on a UCS machine also run.  HUC number 12040101 was run against `gms_pipeline.sh`, then further processed for alpha testing to ensure data integrity.


RobHanna-NOAA:
Updated gms_run_unit.sh and gms_run_branch.sh so all have the '-a' (consistent with gms_pipeline.sh) and assume that the stream orders 1 and 2 will be dropped (by default)

CarsonPruitt-NOAA (approved):

PR #658, opened 2022-08-18
[2pt] PR: Patches for FIM4 alpha test

CarsonPruitt-NOAA:
Patches small bugs in the alpha test functionality.

## Changes

- `src/gms/remove_error_branches.py`: Specified dtype when reading the gms_inputs.csv so HUCs keep their leading zeros.
-  `tools/`
    - `eval_plots.py`: Added prettier FIM labels in the plot legend.
    - `run_test_case.py`: Set remove_inputs to True in mosaic_inundation.
    - `synthesize_test_cases.py`: Made sure that composite FIM 3 test cases are compiled with FIM 4 test cases.

### DIFF
--- a/src/gms/remove_error_branches.py
+++ b/src/gms/remove_error_branches.py
@@ -16,7 +16,7 @@ def remove_error_branches(logfile, gms_inputs):
             print('\nLog file is empty. Skipping this HUC.\n')
             return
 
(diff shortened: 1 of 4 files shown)


mluck (approved):
Tested `remove_error_branches.py` using `gms_run_branch.sh` and generated correct `gms_inputs_removed.csv` output file

Issue #625, opened 2022-07-08
[13pt] FIM outputs into HydroVIS s3 bucket


CarsonPruitt-NOAA:
To make data transfer more smooth, we want to have the API tool upload the files needed for HydroVIS (HAND raster, catchments raster, and hydroTable). The auto-upload should first check the log files to make sure that there were no errors. We should also make sure that the files are visually checked for things that might not throw an error (like excessive file sizes). 

Files needed in s3:
- [ ] HAND raster
- [ ] catchments raster
- [ ] hydroTable.csv

RobHanna-NOAA:
The notes say for the API tool to push it, yet it also asks that it is visually checked. The API tool is all automatted. 
So.. do we:
A) Let the API tool auto push when it is a 'production' run (but if it finds any errors in the (aka.. any that exited with a non zero exit code), then do not push?

B) Do not let the API tool push, but we just run an AWS push from command line. (might have to be from the UCS2 machine.. not sure yet)

Thoughts?

RobHanna-NOAA:
Also.. I have not figured out where to log AWS push error yet or even how to handle that, but will do a 'think' on it.

CarsonPruitt-NOAA:
I think the idea is for the API tool to auto-push the data to S3, and then someone on the HydroVIS side will do a check before they implement the new data in production.

RobHanna-NOAA:
First version u sing Boto3, in a full scale usage (coping all files over to s3) was very slow. It was running successfully but was aborted after 24 hours. That is impractical in a production model.  We will re-design it using CLI and AWS sync api tool. Figuring out dynamic AWS credentials will be the hardest part.

PR #664, opened 2022-08-22
[1pt] Add tool to push data to S3 (for Hydroviz)

RobHanna-NOAA:
A few things are covered in this PR.

1) FIM Outputs to HydroViz `AWS (Amazon Web Service) S3`. While the issue asked for FIM3 to auto push it to `S3`, a decision was made to make it for FIM4 as a stand alone tool.  It can still be integrated easily into the FIM 3 API tool, but the level of effort, considering that the tool is about to be shelved, made it smarter to make this as a stand alone tool for FIM4.  This tool can be run as a command line tool or integrated against other code based. It can also be run on any output including FIM3 output folder.

    This tool has a whitelist system with wildcards, to select which files are sent up to `S3`. A requirement exists to control while files are copied up to S3. While designed and tested for HydroViz, it is configurable to any S3 bucket address and/or folders. While not tested, it is expected to be able to push files and folders to ESIP if required. See [issue 625](https://github.com/NOAA-OWP/inundation-mapping/issues/625).  

2) Adding AWS to the FIM4 required an upgrade to the docker images for AWS Boto3. Piggy-backing on this, we took the opportunity to add three additional packages, `Jupyter`, `JupyterLab` and `Ipympl`. Jupyter is being used by a number of developers and this makes it easier for them to use Jupyter tools with FIM code. Also added bash AWS CLI via the Dockerfile.

3) A minor correction to `gms_run_branch.sh` to move a code block down the page after variable setup and validation.

4) During output processing while `overwriting` past run (gms_pipeline.sh, gms_run_unit.sh, gms_run_branch.sh), the `src_optimization log directory` was not be cleaned. Old logs from previous runs could be carried forward. This is now corrected. See [Issue 661](https://github.com/NOAA-OWP/inundation-mapping/issues/661). 

## Additions

- `src`
   - `data`
       - `README.md`: This system was designed with the expectation that more AWS interaction will be coming in the near future. As a result, a class model using inheritance was created to make future enhancements easier and more standardized. This document talks about the tool and process. Note: The new `data` folder can be used for AWS external communication, but could also be used for other external data communication modules.
       - `aws`
           - `aws_base.py`:  With implementing a new class and inheritance system, this file has properties and a method that all child class will be expected to use and share. This makes it quicker and easier to added new AWS tools and helps keep consistant patterns and standards.
           - `aws_creds_template.env`: There are a number of ways to validate credentials to send data up to S3. We have chosen to use an `.env` file that can be passed into the tool from any location. This is the template for that `.env` file. Later versions may be changed to use AWS profile security system.
           - `s3.py`: This file pushes file and folders up to a defined S3 bucket and root folder. Note: while it is designed only for `puts` (pushing to S3), hooks were added in case functional is added later for `gets` (pull from S3).

## Changes

- `utils`
   - `shared_functions.py`:  A couple of new features
       -  Added a method which accepts a path to a .lst or .txt file with a collection of data and load it into a  python list object. It can be used for a list of HUCS, file paths, or almost anything. 
       - A new method for quick addition of current date/time in output.
       - A new method for quick calculation and formatting of time duration in hours, min and seconds.
       

- `Dockerfile` : Removed a line for reloading Shapely in recent PRs, which for some reason is no longer needed after adding the new BOTO3 python package. Must be related to python packages dependencies. This removed Shapely warning seen as a result of another recent PR. Also added code to load AWS CLI for bash.

- `Pipfile` and `Pipfile.lock`:  Updates for the four new python packages, `boto3` (for AWS), `jupyter`, `jupyterlab` and `ipympl`. We have some staff that use Jupyter in their dev actitivies. Adding this package into the base Docker image will make it easier for them.

## Testing

- `gms_pipeline.sh`, `gms_run_unit.sh` and `gms_run_branch.sh` were tested for a couple HUCs to make sure all python dependencies are still fine.
- Alpha testing was done to ensure all is still well.
- Docker image creation tests were completed
- Tests were done to ensure the new Docker images they auto load Jupyter tools.
- Considerable testing against the new S3 tool including:
   - different files / folder combinations
   - validating that it fails if invalid credentials are passed in.
   - testing of various combinations of files / folder paths and wildcard character usage for the new whitelist.lst file.
   - testing of various combinations (pass and fail) for the CLI input for the `s3.py` file.
- Stub testing with small folders into test S3 buckets
- A full push with whitelist of the 4.0.7.0 folder into the Hydroviz S3 Bucket. (Took 3.5 hours)


RobHanna-NOAA:
Ready for merge. It is pushing a full BED run of 4.0.7.0 from previous_fim folder up to Hydroviz (with whitelist filtering) as we speak and is doing just fine. I will check in the morning to see how it finished up.

RobHanna-NOAA:
And I updated the code to move the code up one level as we discussed earlier today.

RobHanna-NOAA:
Now updated to use bash / AWS CLI instead of Boto3. Boto3 was taking 24 hours plus to push all of the files up to S3, but aws cli can do it in 3.5 hours. With AWS CLI, it is not as easy to track and catch errors, but Boto3 does not yet have a bulk (recursive folders) tool and even with adding its performance upgrades, it is not good to be reasonable.

This is again ready for merging.  New Docker images will need to be created after merge.

gms_run_branch.sh
```diff
@@ -176,6 +161,21 @@ elif [ $overwrite -eq 1 ]; then
...
+## CONNECT TO CALIBRATION POSTGRESQL DATABASE (OPTIONAL) ##
+if [ "$src_adjust_spatial" = "True" ]; then
+    if [ ! -f $CALB_DB_KEYS_FILE ]; then
+        echo "ERROR! - src_adjust_spatial parameter is set to "True" (see parameter file), but the provided calibration database access keys file does not exist: $CALB_DB_KEYS_FILE"
+        exit 1
+    else
+        source $CALB_DB_KEYS_FILE
+        echo "Populate PostgrSQL database with benchmark FIM extent points and HUC attributes"
+        echo "Loading HUC Data"
+        time ogr2ogr -overwrite -nln hucs -a_srs ESRI:102039 -f PostgreSQL PG:"host=$CALIBRATION_DB_HOST dbname=calibration user=$CALIBRATION_DB_USER_NAME password=$CALIBRATION_DB_PASS" $input_WBD_gdb WBDHU8
```
CarsonPruitt-NOAA, line 173: In `dev`, we changed `$input_WBD_gdb` to be `$inputDataDir/wbd/WBD_National.gpkg` because that variable wasn't coming through.

Dockerfile
```diff
@@ -87,7 +92,7 @@ ENV PYTHONPATH=${PYTHONPATH}:$srcDir:$projectDir/tests:$projectDir/tools
 COPY Pipfile .
 COPY Pipfile.lock .
 RUN pip3 install pipenv==2022.4.8 && PIP_NO_CACHE_DIR=off PIP_NO_BINARY=shapely,pygeos pipenv install --system --deploy --ignore-pipfile
-RUN pip3 install shapely
+## RUN pip3 install shapely
```
CarsonPruitt-NOAA, line 95: Has this been tested? If leaving out the shapely install works, let's just remove this line altogether instead of keeping the commented line in here without any notes as to why.

gms_run_branch.sh
RobHanna-NOAA: oh thats fine. I was just trying to move the whole block.  I will take a look at it.

Dockerfile
RobHanna-NOAA: can do

CarsonPruitt-NOAA (approved):

Issue #660, opened 2022-08-19
[13pt] Derive_level_paths.py will choose smaller order streams if the arbolate sum is higher


CarsonPruitt-NOAA:
Derive_level_paths.py is agnostic to stream orders and will choose the wrong level path when it only considers arbolate sum. Higher order streams will sometimes have lower arbolate sums at confluences near HUC8 boundaries.

An example is 12040104 near the AHPS site SHLT2. The level path for branch `1440000002` veers off the main river and assigns a new level path '1440000004' near the upstream HUC8 boundary.
![image](https://user-images.githubusercontent.com/90792257/185669238-b1d3f7f2-db91-4935-a21c-fb896d1856a1.png)

PR #672, opened 2022-08-26
[1pt] PR: Update derive level paths to use stream order

CarsonPruitt-NOAA:
When deriving level paths using only arbolate sum, branches end up veering off the main river near the upstream HUC8 boundary because the main river has a lower arbolate sum due to being cut off at the WBD. This improvement uses stream order to override that behavior so that the branch stays on the main river. Addresses #660.

## Changes

- `src/gms/`
    - `derive_level_paths.py`: Added `order_` to the list of comparison_attributes in the `derive_stream_branches` method.
    - `stream_branches.py`: Updated the `derive_stream_branches` method to first consider stream order when choosing which upstream segment to continue the current level path. If both upstream segment have the same order, then arbolate sum is used.
-  `unit_tests/gms/derive_level_paths_unittests.py`: Updated the import to the new class.

## Testing

1. This PR has been tested on the benchmark hucs using the following command:
```
gms_pipeline.sh -u /data/inputs/huc_lists/all_eval_hucs_20220506.lst -n carson_order_test -j 80
```

## Screenshots
BEFORE: The main branch (blue) veers off to the northern tributary because it has a higher arbolate sum.
image

AFTER: The main branch (blue) continues on the larger river because its higher stream order now overrides the lower arbolate sum.
image

The issues we were seeing with `shlt2` are fixed by this PR. (see screenshot in #660)
image

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,12 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## 4.0.9.0 - 2022-09-09 - [PR #672](https://github.com/NOAA-OWP/inundation-mapping/pull/672)
+
+When deriving level paths, this improvement allows stream order to override arbolate sum when selecting the proper upstream segment to continue the current branch.
+
+<br/><br/>
+
 ## 4.0.8.0 - 2022-08-26 - [PR #671](https://github.com/NOAA-OWP/inundation-mapping/pull/671)
 
 Trims ends of branches that are in waterbodies; also removes branches if they are entirely in a waterbody.
--- a/src/gms/derive_level_paths.py
+++ b/src/gms/derive_level_paths.py
@@ -96,7 +96,7 @@ def Derive_level_paths(in_stream_network, out_stream_network, branch_id_attribut
                                                             upstreams=upstreams,
                                                             branch_id_attribute=branch_id_attribute,
                                                             reach_id_attribute=reach_id_attribute,
-                                                            comparison_attributes='arbolate_sum',
+                                                            comparison_attributes=['arbolate_sum', 'order_'],
                                                             comparison_function=max,
                                                             verbose=verbose
                                                            )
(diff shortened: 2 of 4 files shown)


CarsonPruitt-NOAA:
@mluck I'd like you to do a very thorough review of this to make sure that we're not introducing any unintended changes to the level paths. Here's some specific places I want to ensure have good behavior:

1. Multi-channel areas. I think that `02020005` might have an example of this. It might be good to look at the NWM flowlines layer to find another HUC with multiple channels.
2. Confluences where the two streams have the same order. These confluences should have the same behavior before and after this change (level path should be continued on the higher arbolate sum segment).

mluck:
In the confluences of same orders, the correct branch is selected by arbolate sum; however, this seems to have a problem in the multi-channel areas -- compare the lower left segments in the images of branches to stream orders below, flow is from lower left to upper right -- the first order channel is propagated upstream:

Stream order (by line width)
image

Branches
image

mluck:
The attribute information is correct where the flow separates -- might be possible to use that?

image

CarsonPruitt-NOAA:
I think what's going is that the main branch (long purple in your screenshot) is correctly being propagated upstream, but then the incorrect (pink) branch is put in the queue. When the pink branch eventually gets run, it's overwriting the purple branch ID in all of the upstream segments because it connects back into it. Do you think that's an accurate assessment?

If that's the case, I'm not sure it's feasible to fix that behavior without introducing more problems. Since it's not a fatal error, we may just have to keep this kind of behavior in mind if we ever run this at full-GMS.

mluck:
Here's a closer look at what's going on with the data in `02020005/branches/`: the upstream branch (3246000213) follows A-B, while the mainstem branch (3246000001) follows C-D in nwm_streams_subset_levelPaths and B-D in demDerived_reaches_split_filtered_addedAttributes. According to nwm_flows, A-C-D is 4th order and B is first order.
image

CarsonPruitt-NOAA:
There are a few test cases where the CSI decreases. I've looked into most of these cases and can confidently say that these are caused by differences in catchment boundaries between the two versions, not because of a decrease in skill.
image
image

mluck (approved):
Checked this on several HUCs. Looks good to me.

PR #679, opened 2022-09-13
[1pt] PR: Fix thalweg notch

### LINKED ISSUE #673: [8pt] Fix reverse flow issue
We want to fix the reversed flow issue caused by the longitudinal thalweg elevation conditioning. The root cause of the issue is NHD meandering in and out of the buffered HUC polygon as seen in the screenshot. The relevant datasets are: `NHDPlusBurnLineEvent_subset.gpkg` and `wbd_buffered.gpkg`.

image

mluck:
Fixes thalweg notch created by clipping upstream ends of the stream segments to prevent the stream network from reaching the edge of the DEM and being treated as outlets when pit filling the burned DEM (closes #673)

## Changes

- `src/clip_vectors_to_wbd.py`: Uses a slightly smaller buffer than wbd_buffer (`wbd_buffer_distance-2*[DEM cell size]`) to clip stream network inside of DEM extent. 

## Testing

1. Ran `fim_run.sh` on HUCs 03010203 03110205 08030208 08030100 10030203 11030018 12110111 16050103 18010203 18010209.


CarsonPruitt-NOAA:
Can you check the following HUCs: `12060201, 04300202, 13040201, 07070003`.
They are still showing some deep ~100m thalweg notches and we need to make sure that the "meandering flowlines" that we're trying to fix aren't the issue here.

mluck:
There are a few different issues here: 

1. This seems to be an issue that the smaller buffer (used to clip the stream lines (blue) before rasterizing the stream network (black)) wasn't quite small enough to keep the rasterized network from reaching the DEM edge even though we used 2 x cellsize (20 m) to account for the cell diagonal. Decreasing the inner buffer by another 10 m (3 x cellsize) should fix this problem. Here are some examples:

07070003
image


12060201
image

The other two HUCs didn't have stream network rasters reaching the edge of the DEM but had other problems that are still being investigated:

2. HUC 13040201 crashed with a `Killed` message that I've seen happen before when Docker runs out of memory
```
##########################################################################
Preprocessing for lateral thalweg adjustment 13040201
##########################################################################
Tue Sep 20 13:48:56 UTC 2022
/foss_fim/src/run_by_unit.sh: line 186:  4021 Killed                  python3 -m memory_profiler $srcDir/unique_pixel_and_allocation.py -s $outputHucDataDir/demDerived_streamPixels.tif -o $outputHucDataDir/demDerived_streamPixels_ids.tif -g $outputHucDataDir/temp_grass
Command exited with non-zero status 137
	Command being timed: "/foss_fim/src/run_by_unit.sh 13040201"
	User time (seconds): 747.85
	System time (seconds): 105.14
	Percent of CPU this job got: 52%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 26:53.63
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 9600212
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 1984
	Minor (reclaiming a frame) page faults: 15805828
	Voluntary context switches: 72117
	Involuntary context switches: 272739
	Swaps: 0
	File system inputs: 26716528
	File system outputs: 52258816
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 137
```

3.  HUC 04300202: This one has an edge of a tile visible in the raw DEM (gray scale). The pixels in the tile edge have an elevation much lower than the surrounding terrain (~150 m below the neighboring pixels at the stream crossing). I'm not sure yet if this is causing the problem but it does cause some funky things where the tile edge cuts across `rem_zeroed_masked.tif` (colored)
image

image

The tile edge might be confusing the dem burning process:
raw DEM (`dem_meters.tif`) vs burned DEM (`dem_burned.tif`)
image
image

mluck:
1. Changing the stream network buffer to (wbd_buffer - 3*cellsize) fixed the issue in 07070003 and 12060201. The remaining issues below seem to be unrelated to the stream network reaching the edge of the DEM.
2. 13040201 finished this time but looks like the thalweg notch is still there -- possibly because the streamline (blue) does not fall in the darker channel (the streamline is actually the US-Mexico border) as shown below:
image
3. The tile edge issue in 04300202 remains.

CarsonPruitt-NOAA:
Thanks for looking into these other HUCs. I think we've solved the "meandering stream" issue that was originally identified, so this PR is probably good to be merged in after we compare alpha test results.

The tile edge issue seems like something that would be difficult to fix outside of updating out DEMs. Hopefully the 3DEP update in FIM v4.X will solve this.

RobHanna-NOAA:
Rob Testing:
Did a full alpha (216 hucs) run for both FR and MS, then an alpha test composite and eval plots. They looked just fine and were identical to 3.0.34.2

PR #696, opened 2022-10-03
[1pt] fixes for calib system, landSea deny list and params

RobHanna-NOAA:
Comment out LandSea_subset.gpkg in deny_gms_unit_prod.lst so it does not get deleted. Some branches need it for processing.

Also.. change default for params_template.env -> src_adjust_spatial="False", back to default of "True"

During fixing this issue, Ryan and I found a couple of others related to the calib system and are now fixed in this PR.

Ryan helped fix src_adjust_usgs_rating.py and src_adjust_usgs_rating.py. It was discovered if you are processing just one huc that is not in the calibration database, it would fail (yes. some hucs will not be in the db). We also discovered and fixed an infinite loop where it attempts to call the calib database but can not connect. Now, it will make 6 attempts to connect and will abort attempts to connect (instead of indefinitely trying to connect.

## Changes

-

## Testing

To test the HUC not in the calib db, we used HUC 02060001 which does not/will not have data in the calib db.  It can also be used to test the missing LandSea file.  When you run this with this fix, in the unit folder, you will see the LandSea_subset.gpkg now exists, and in fim 4.0.9.4, this file would be missing.

To test the fix for the infinite loop, we commented out the code in gms_run_branch.sh for the code block "## CONNECT TO CALIBRATION POSTGRESQL DATABASE (OPTIONAL) ##" as a quick failed database connection emulator, then ran it.  There are multiple ways to emulate the db not being available, this being one. 

Also.. by default, the calib system should now fully run as expected.

### DIFF
--- a/config/deny_gms_unit_prod.lst
+++ b/config/deny_gms_unit_prod.lst
@@ -3,7 +3,7 @@
 NHDPlusBurnLineEvent_subset.gpkg
 #branch_id.lst
 #branch_polygons.gpkg
-LandSea_subset.gpkg
+#LandSea_subset.gpkg
 nhd_headwater_points_subset.gpkg
 #nld_subset_levees.gpkg
 #nwm_catchments_proj_subset.gpkg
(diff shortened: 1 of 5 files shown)


RobHanna-NOAA:
Ready for review, approval and merge.

RyanSpies-NOAA (approved):
Successfully tested this feature branch on two hucs (02060001: no calibration data; 12030102: USGS & spatial calibration). Approved for merge

CarsonPruitt-NOAA (approved):

PR #700, opened 2022-10-05
[1pt] PR: Calibration DB Docker Compose Update

nickchadwick-noaa:
A small change made to the Calibration DB's `docker-compose.yml` file to fix a couple of issues that were causing the DB data to not persist after the host machine was rebooted.

This has already been tested on the dev and prod machines for a while and seems to have solved the issue.


RobHanna-NOAA (approved):

RobHanna-NOAA:
Why is the actual pgdata folder in the var directory instead of a more visible mount (say.. dev_fim_share/foss_fim/calibration/data/pgdata)  Just curious?    But.. the PR is just fine and I was able to get it running on Dev2

nickchadwick-noaa:
There is a long-standing issue with Docker where host mounts (a docker volume attached to an actual folder on the host filesystem) is not immediately available upon reboot of the host system. So the issue was happening on the weekly IT patching and rebooting that after the host was up and docker-compose brought back up the containers that were running, it was not properly mounting the PGDATA mount.

To remedy this, I switched the PGDATA from a host mount to a regular named Docker volume, which does not suffer from this issue.

You can ready more about this issue here https://github.com/docker/for-win/issues/584#issuecomment-286792858

CarsonPruitt-NOAA (approved):

PR #707, opened 2022-10-11
[1pt] Adds capability to produce single FIM3 rating curve comparison plots for each gage

mluck:
Adds capability to produce single rating curve comparison plots for each gage. This PR is identical to #709 except this one is for FIM3.

## Additions

- `tools/rating_curve_comparison.py`
    - Adds `generate_single_plot()` to make a single rating curve comparison plot for each gage in a given HUC
    - Adds command line switch to generate single plots

### DIFF
--- a/tools/rating_curve_comparison.py
+++ b/tools/rating_curve_comparison.py
@@ -236,11 +236,10 @@ def generate_rating_curve_metrics(args):
         # plot rating curves
         if alt_plot:
             generate_rc_and_rem_plots(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table, hydrotable_filename)
+        elif single_plot:
+            generate_single_plot(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table)
         else:
             generate_facet_plot(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table)
-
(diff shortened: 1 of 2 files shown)


tools/rating_curve_comparison.py
```diff
@@ -238,6 +239,9 @@ def generate_rating_curve_metrics(args):
         else:
             generate_facet_plot(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table)
 
+        if single_plot:
```
CarsonPruitt-NOAA, line 242: Instead of doing it this way where it will produce both facet plots and single gage plots, I think we should do single plots OR the default facet plots. Something like this:
```
        # plot rating curves
        if alt_plot:
            generate_rc_and_rem_plots(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table, hydrotable_filename)
        elif single_plot:
            generate_single_plot(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table)
        else:
            generate_facet_plot(rating_curves, rc_comparison_plot_filename, nwm_recurr_data_table)

```

CarsonPruitt-NOAA (approved):
Tested on full BED and works great.

PR #708, opened 2022-10-11
[1pt] PR: Cleanup Branch Zero outputs

RobHanna-NOAA:
Change branch zero output cleanups to no longer be done after processing units. An recent PR, changed gms_run_branch for a second branch zero output cleanup after all branches have been processed. This means the first cleanup of branch zero is no longer needed.  Branch zero is cleaned at the end as some branches need some of those files during their processing.

### Removals

- `config`
    - `deny_gms_branch_zero.lst`:  no longer needed as branch zero will be cleaned the same way all other branches are cleaned using other deny lists.

### Changes

- `gms_pipeline.sh`: A quick help text update.
- `gms_run_branch.sh`: A couple of quick text updates.
- `gms_run_unit.sh`: A quick help text update.
- `src`
    - `gms`
        - `gms_run_unit.sh`: Remove the block that does an pre-branch processing, branch zero cleanup.

### Testing

tested gms_pipeline with four permutations:
1) use defaults for deny list for units and branches
2) change the -ud (deny unit) flag to the word "none"), meaning do not cleanup unit output files.
3) change the -bd (deny branch flag to the word "none"), meaning do not cleanup branch output files.
4) override both the -ud and -bd flags against customized versions of both the  deny unit files and the deny branch files.


RobHanna-NOAA:
Cancelled. Discovered that we have to keep some cleanup for branch zero as it is now. Fundamentally resulting it in being cleaned up twice. But... we have to keep both sets of cleanup. Without the initial cleanup, it uses up to much disk space.

PR #718, opened 2022-10-19
[1pt] PR: Port thalweg notch fix from FIM3

mluck:
Ports thalweg notch fix from FIM3 (#679: dev-fim3-fix-reverse-flow), which fixes thalweg notch created by clipping upstream ends of the stream segments to prevent the stream network from reaching the edge of the DEM and being treated as outlets when pit filling the burned DEM.

## Changes
- `src/clip_vectors_to_wbd.py`: Uses a slightly smaller buffer than wbd_buffer (wbd_buffer_distance-3*[DEM cell size]) to clip stream network inside of DEM extent.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,16 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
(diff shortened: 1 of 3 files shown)


RobHanna-NOAA (dismissed):
Looks Good

CarsonPruitt-NOAA (approved):

PR #723, opened 2022-10-24
[1pt] PR: Runs branch 0 on HUCs without other branches after stream order filtering

### LINKED ISSUE #715: [8pt] Fix Units with only branch zero (no level paths)
14 HUCs exit at the unit processing level with an exit code of 60, meaning no stream orders remaining that are level 3 and higher. However.. we still need to process branch zero to fill in the HUC.  This will result in only branch zero and no other branches for that HUC.

We need to trap the error, suppress it, no need for logging it, and continue to branch zero processing for it.

HUCS affected: 01010011, 02030201, 03130014, 04020203, 10050003, 10060007, 12040203, 12080001, 12100403, 12110203, 12110207, 17110003, 18070201, 20030000.

mluck:
Runs branch 0 on HUCs with no other branches remaining after filtering stream orders if `drop_low_stream_orders` is used. Closes #715.

## Additions

- `src/gms`
    - `stream_branches.py`: adds `exclude_attribute_values()` to filter out stream orders 1&2 outside of `load_file()`

## Changes

- `src/gms`
    - `buffer_stream_branches.py`: adds check for `streams_file`
    - `derive_level_paths.py`: checks length of `stream_network` before filtering out stream orders 1&2, then filters using `stream_network.exclude_attribute_values()`
    - `generate_branch_list.py`: adds check for `stream_network_dissolved`

## Testing

1. Ran `gms_run_unit.sh` and `gms_run_branch.sh` on HUCs 10060007 (no 3+ stream orders) and 12030102

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,24 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v4.[to be assigned] - 2022-10-24 - [PR #723](https://github.com/NOAA-OWP/inundation-mapping/pull/723)
+
+Runs branch 0 on HUCs with no other branches remaining after filtering stream orders if `drop_low_stream_orders` is used.
+
+### Additions
+
+- `src/gms`
+    - `stream_branches.py`: adds `exclude_attribute_values()` to filter out stream orders 1&2 outside of `load_file()`
+
+### Changes
+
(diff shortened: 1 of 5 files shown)


RobHanna-NOAA (dismissed):
Looks good. I did a set of five, three of them were known to fail earlier with no remaining stream orders, and two that normally.  They all processed fine and the outputs look good.

CarsonPruitt-NOAA (approved):

PR #760, opened 2022-12-13
[8pt] PR: Stacked Bar Eval Plots

CarsonPruitt-NOAA:
This PR adds the ability to produce the stacked bar eval plots using `/tools/eval_plots_stackedbar.py`. These plots are a subjective alternative to using summary statistics (e.g. CSI. FAR, MOD, or MCC) and are a good way to compare under/over predictive tendencies between flows or versions.

## Additions

- `/tools/eval_plots_stackedbar.py`: produces stacked bar eval plots in the same manner as `eval_plots.py`.

## Notes

 - The data displayed in these plots are calculated by taking the cell counts of each contingency category and normalizing by the sum of the True Positive and False Negative counts. By normalizing in this manner and stacking the contingency category bars on each other, it's possible to discern a relative underprediction by looking at the yellow bar and relative overprediction by looking at the length of the red bar above 1. See the first screenshot for a graphical representation.
 - Since this tool uses the same output file structure as the normal `eval_plots.py`, you can use the same workspace `-w` argument for both commands to produce both types of plots.
 - The `-i` flag produces individual plots for each benchmark site.
 - Since this tool puts all individual plots into one figure, it's possible to surpass the matplotlib limit for figure size if the user inputs too many versions.
 - Use the `-iqr` flag to display error bars that indicated the inter-quartile range for each flood threshold.
 - The `-d` flag produces difference plots based on the supplied metric (CSI, MCC, FAR, or PND)

## Sample Usage

`python /foss_fim/tools/eval_plots_stackedbar.py -m /data/previous_fim/fim_4_0_13_1/fim_4_0_13_1_metrics.csv -w /data/temp/carson/fim4_alpha/v4_0_13_1 -v fim_1_0_0 fim_2_x_comp fim_3_0_36_0_comp_c fim_4_0_0_0 fim_4_0_13_1`

`python /foss_fim/tools/eval_plots_stackedbar.py -m /data/previous_fim/fim_4_0_13_1/fim_4_0_13_1_metrics.csv -w /data/temp/carson/fim4_alpha/v4_0_13_1 -v fim_4_0_0_0 fim_4_0_13_1 -d CSI`

## Screenshots

Explanation of where the data comes from. See the notes section for a detailed description.
image

Aggregate plots (default behavior)
image

Individial Plots (with the `-i` flag)
image

Diff plots (with the `-d` flag)
image


tools/eval_plots_stackedbar.py
```diff
@@ -0,0 +1,324 @@
...
+    parser.add_argument('-w', '--workspace', help = 'Output workspace', required = True)
+    parser.add_argument('-v', '--versions', help = 'List of versions to be plotted/aggregated. Versions are filtered using the "startswith" approach. For example, ["fim_","fb1"] would retain all versions that began with "fim_" (e.g. fim_1..., fim_2..., fim_3...) as well as any feature branch that began with "fb". An other example ["fim_3","fb"] would result in all fim_3 versions being plotted along with the fb.', nargs = '+', default = [])
+    parser.add_argument('-i', '--site-plots', help = 'If enabled individual barplots for each site are created.', action = 'store_true', required = False)
+    parser.add_argument('-iqr', '--show-iqr', help = 'If enabled, inter-quartile range error bars will be added.', action = 'store_true', required = False)
+    parser.add_argument('-d', '--diff-stat', help = 'This option creates diff plots instead of stacked bar plots. Only 2 versions can be used with this option. ' + \
+        "Input one of the following statistics to be used for comparison: ('CSI', 'MCC', 'TPR', 'PND', 'FAR').", default = False, required = True)
```
RyanSpies-NOAA, line 308: @CarsonPruitt-NOAA I believe this argument should be "optional". Throws an error if you leave it out. I made the change and tested it successfully.

CarsonPruitt-NOAA: Ah good catch... go ahead and push your fix if you've already coded it up.

RyanSpies-NOAA (approved):
Tested the code with `-d` flag, -`i`, and without `-d` flag. All outputs plots look good. Future suggestion: the large inidividual plots can occasionally throw an error: `ValueError: Image size of 800x72125 pixels is too large. It must be less than 2^16 in each direction.` Suggest finding a way to split those by site/huc/etc.

PR #767, opened 2022-12-20
[1pt] PR: Bug fix: `inundation.py` ignoring nodata

### LINKED ISSUE #761: [8pt] Inundation of masked areas of REM
Nodata areas of the relative elevation model (REM) are showing up as inundated rather than being masked.

The nodata value (e.g., -9999) of the REM isn't being masked but instead is read as a value by `tools/inundation.py`, which results in that value being added to the inundation stage to produce extremely high depth values which are always inundated.

Although this issue may have existed for a while, it hasn't been apparent until the levee-protected areas were masked from the REM.

mluck:
Bug fix for `inundation.py` ignoring nodata value and inundating masked areas in REM raster. Fixes #761

## Changes

- `tools/inundation.py`: Sets depth to `0` if REM value is less than `0`

## Testing

1. Ran `inundation.py` on HUC 12030109 branch 0. Verified values in QGIS (screenshot below).
2. Ran `gms_pipeline.sh` and `synthesize_test_cases.py` on HUC 12030109 without errors.

## Screenshots

Levee-protected areas (gray, 50% transparent) are no being inundated, for example in the depth raster:
image

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,16 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v4.(to be assigned) - 2022-12-20 - [PR #767](https://github.com/NOAA-OWP/inundation-mapping/pull/767)
+
+Fixes inundation of nodata areas of REM.
+
+### Changes
+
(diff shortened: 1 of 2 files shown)


RobHanna-NOAA (dismissed):
Looks good. I ran some hucs and tested inundation on them to see improvement in inundation around a levee protected areas and it was more accurate with the fix.

PR #782, opened 2023-01-06
[1pt] PR: Change the projection of HAND processing to EPSG:5070

CarsonPruitt-NOAA:
Changes the projection of HAND processing to EPSG:5070.

### Changes

- `gms_run_post_processing.sh`: Adds target projection for `points`
- `data/nld/preprocess_levee_protected_areas.py`: Changed to use `utils.shared_variables.DEFAULT_FIM_PROJECTION_CRS`
- `src/`
    - `src_adjust_spatial_obs.py`: Changed to use `utils.shared_variables.DEFAULT_FIM_PROJECTION_CRS`
    - `utils/shared_variables.py`: Changes the designated projection variables
    - `gms/`
        - `stream_branches.py`: Checks the projection of the input streams and changes if necessary
        - `run_by_unit.py`: Changes the default projection crs variable and added as HUC target projection
- `tools/inundate_nation.py`: Changed to use `utils.shared_variables.PREP_PROJECTION`

## Testing

1. Tested this feature branch on the benchmark HUCs and found zero differences between its scores and v4.0.12.1 (the version off which this one branched).

## Notes

- In order to be complete, we should also change the CRS of all of our input data. As of now, only the 3DEP DEMs are in this new projection.


mluck (approved):
It may be a good idea to make sure input datasets are in 5070 as well to minimize reprojecting during runs.

RobHanna-NOAA:
Stand by.. I think we need to talk about this as their could be more code changes and more tests.

RobHanna-NOAA:
A couple of things: We should do a code search for the phrases "1020309" and "4269". Not confirmed but we might need an adjustment in clip_vectors_to_wbd.py  (levee_protected_areas = levee_protected_areas.to_crs('ESRI:102039').  There are some other places that those two CRS's should be at least double checked.

I also recommend looking at src_adjust_spatial_obs.py  (appx line 160 for water_edge_df = ... This into itself is not an issue but the fact that we had to force it suggest there might be a hidden crs problem somewhere else (aka.. why did he need it in the first place).

I also wonder if we should double check all input data (jsut a few from each file type), so we know exactly the CRS coming in, to make sure we are reprojection to 5070.  Aka.. is any slipping through that the input source is something like 4269 and we don't re-project it?  Maybe I am being overly cautious but there is some evidence that a deeper review is required.

mluck:
In order for this PR to succeed, the calibration database will need to be rebuilt to reload `points` in EPSG:5070

RobHanna-NOAA:
@RyanSpies-NOAA  Can you look at the reloading of the calib db issue?  I can do the full testing of the PR once that has been figured out.  I am also not 100% sure how to reload that db. I will look at the readme in our code to see if it just kills the connection or actually deletes and rebuilds the database when the calib postgres service is stopped.

mluck:
To reload the calibration database on my EC2, I did this (from https://github.com/NOAA-OWP/inundation-mapping/tree/dev/tools/calibration-db#destroying-the-database-from-the-toolscalibration-db-directory):
```
docker-compose down
rm -rf pgdata
```
Then redo the `docker-compose --env-file {path to .env file}/{your key file name} up --build -d` step.

RyanSpies-NOAA:
For the calb database projection handling. I think we can make some simple revisions to the ogr2ogr lines 144 and 146 in https://github.com/NOAA-OWP/inundation-mapping/blob/dev/gms_run_post_processing.sh

RobHanna-NOAA:
Ryan.. I think he has the changes in his PR for changing those lines. Is the file changes in this PR not ok?

RobHanna-NOAA:
oh ya.. those lines depend on the CRS of the WBD_National.gpkg.  hummm.. ya.. we need a quick meeting on this one

RobHanna-NOAA:
Yup.. I get it.. I think we are just fine now. Thanks for humoring me and making sure all is well.

RobHanna-NOAA:
Not wild about having all of the output files having one of two CRS's (102039 or 5070). I would prefer all outputs were 5070, but the three key output files of the demDerived_reaches_split_filtered_addedAttributes_crosswalked_{}.gpkg, gw_catchments_reaches_filtered_addedAttributes_crosswalked_{}.gpkg and rem_zeroed_masked_{}.tif are all at 5070. I did a bunch of tests including raster math on the pre PR (102039) and post PR (5070) for the rem and they were completely identical. So.. There is no harm on a mixed CRS file outputs, just need to be super careful. Easy to make mistakes.  I strongly recommend we get the inputs files updated and changed to CRS 5070 as per issue 682.

RobHanna-NOAA (approved):

RobHanna-NOAA (approved):

RobHanna-NOAA (approved):

PR #744, opened 2022-11-16
[3pt] PR: Replace NHD headwaters with NWM-derived headwaters and remove NHD references

mluck:
Changes branch 0 headwaters data source from NHD to NWS to be consistent with level path branches. Removes references to NHD flowlines and headwater data. Closes #742.

## Changes

- `src/gms/derive_level_paths.py`: Generates headwaters before stream branch filtering

## Removals

- Removes NHD flowlines and headwater references from `gms_run_unit.sh`, `config/deny_gms_unit_prod.lst`, `src/clip_vectors_to_wbd.py`, `src/gms/run_by_unit.sh`, `unit_tests/__template_unittests.py`, `unit_tests/clip_vectors_to_wbd_params.json`, and `unit_tests/clip_vectors_to_wbd_unittests.py`

## Testing

1. BLE benchmark CSIs compared on HUCs 11110202 and 12090301

HUC8 | Magnitude (yr) | dev | dev-cleanup-nhd
-- | -- | -- | --
11110202 | 100 | 0.643 | 0.638
11110202 | 500 | 0.651 | 0.645
12090301 | 100 | 0.666 | 0.663
12090301 | 500 | 0.690 | 0.690

## Screenshots
REM grid differences between `dev` and `dev-cleanup-nhd` (grayscale from -10 to 10 m). Differences exist especially where the NHD and NWM headwater points ended up in different catchments, but overall the REM difference is largely zero.

1. 11110202
image

2. 12090301
image

PR #790, opened 2023-01-11
[1pt] PR: Remove Great Lakes clipping

### LINKED ISSUE #783: [5pt] WBD clip for Great Lake HUCs
The clipping of the WBD layer for HUCs on the Great Lakes may not be necessary anymore as we've changed the input streams to use the NWM network. 

[Here's the section of code where this occurs](https://github.com/NOAA-OWP/inundation-mapping/blob/e92eb8dbe6f366f98ca741b4f76c49b5516a5834/src/clip_vectors_to_wbd.py#L53)

mluck:
Removes Great Lakes clipping. Unnecessary since removal of NHD datasets (PR #744). Closes #783.

## Changes

- `src/`
    - `clip_vectors_to_wbd.py`: Removed Great Lakes clipping section and references to Great Lakes polygon dataset.
    - `gms/run_by_unit.sh`: Removed parameters for Great Lakes polygons and lake buffer size

## Testing

1. `gms_pipeline.sh` produced equivalent results on the original problem HUCs (04040002 and 04060101) with Great Lakes clipping removed.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,19 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## v4.X.XX.X - 2023-01-11 - [PR#790](https://github.com/NOAA-OWP/inundation-mapping/pull/790)
+
+Remove Great Lakes clipping
+
+### Changes
+
+- `src/`
+    - `clip_vectors_to_wbd.py`: Removes Great Lakes clipping and references to Great Lakes polygons and lake buffer size
+
+    - `gms/run_by_unit.sh`: Removes Great Lakes polygon and lake buffer size arguments to `src/clip_vectors_to_wbd.py`
+
+<br/><br/>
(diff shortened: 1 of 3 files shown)


RobHanna-NOAA:
Note: We appear to still use `$input_GL_boundaries = $inputDataDir/landsea/gl_water_polygons.gpkg` (via gms_run_unit.sh), which gets propagated to gms/run_by_unit.sh as $input_GL_boundaries.   As a result, if the HUC starts with 04, then it comes in via the input_LANDSEA variable into clip_vectors_to_wbd.py.  All is fine, we just have to make sure we don't drop that input data.

RobHanna-NOAA (approved):
Looks good. We now get a few streams that extend out into the gl but most are just by under 100 meters and don't appear to hurt anything. I compared streams from HUC's 04010101 04050002 04060104 and found a handful with the extra stream length (compared to fim_4_0_18_0). I don't think it impacts anything.

CarsonPruitt-NOAA (approved):

PR #802, opened 2023-01-24
[1pt] PR: branch trimming for lake segments

RyanSpies-NOAA:
This revision includes a slight alteration to the filtering technique used to trim/remove lakeid nwm_reaches that exist at the upstream end of each branch network. By keeping a single lakeid reach at the branch level, we can avoid issues with the branch headwater point starting at a lake boundary. This ensures the headwater catchments for some branches are properly identified as a lake catchment (no inundation produced). 

## Changes

- `src/gms/stream_branches.py`: New changes to the `find_upstream_reaches_in_waterbodies` function: Added a step to create a list of nonlake segments (lakeid = -9999) . Use the list of nonlake reaches to allow the filter to keep a the first lakeid reach that connects to a nonlake segment.

## Testing

1. Feature branch was tested for the test case domain (200+ hucs). Evaluation results available: `/temp/ryan/rob_ryan_upbrancheval2/`

## Screenshots
List of HUCs tested
image

Solution:
image

Illustrations of inundation changes:
image
image
image

## Notes
Confirmed that this change has a minimal impact on the alpha evaluation metrics:
usgs_comp_stackedbar
ble_comp_stackedbar

## Todos

- This feature branch does not resolve all cases of NWM water body inundation. While somewhat rare, there are examples of unintended inundation occurring when a water body is located off-channel but near a substantial river. Most of these situations seem to occur on relatively small water bodies that are likely to be inundated by the nearby river; however, this deviates from our current practice of avoiding inundation in any of the NWM water bodies. 
Example from HUC 11070205
image


mluck (approved):
Nice work here.

Note that although this PR correctly fixes the issue (lake inundation problem at the head of level paths), the CSI may decrease slightly as inundation exceeding the lake mask margin in the pre-PR version is changed to false negative, and the reduction in CSI will be greater at higher magnitudes (moderate magnitude shown).

Before (CSI=0.54007):
image

After (CSI=0.53965):
image

RyanSpies-NOAA:
Good catch Matt - this is also a helpful reminder that the masking in the alpha evaluation only masks the NWM water body polygon but the underlying benchmark FIM extent includes catchment areas that we've "excluded" from the simulated FIM (unfairly scoring False Negatives).

mluck:
Good point -- would it make sense to also mask excluded catchments from evaluation?

RyanSpies-NOAA:
I agree - ideally we could mask the excluded catchment areas but that may be overly challenging with the overlapping GMS branch catchments. I'm not sure we even want to use the hydroid catchments as a mask because they can easily change between different FIM versions. Brad and I talked a while back about possibly using the NWM catchments (identify the lake catchments) to have a consistent mask input to the alpha eval, but we weren't sure if it was worth the effort at the time.

CarsonPruitt-NOAA (approved):

RyanSpies-NOAA:
Additional note:
image

PR #812, opened 2023-02-14
[1pt] Update for calib db pathing to work in AWS

RobHanna-NOAA:
An update was required to adjust host name when in the AWS environment

### Changes

- `fim_post_processing.sh`: Added an "if isAWS" flag system based on the input command args from fim_pipeline.sh or 

- `tools/calibration-db`
    - `README.md`: Minor text correction.

### Testing

1. Tested in an EC2 instance (with IsAWS = True)
2.  Tested in NWC servers with IsAWS = false.


robgpita-noaa (approved):
Looks good to me! Pretty simple fix. One question I have is the use of the `-s` flag for `curl`. Did you want the output "silent" or was the `-s` a holdover from using `hostname`?

RobHanna-NOAA:
Its is "-s" for silent. Without it, we get a ton of info back and not just the name.

CarsonPruitt-NOAA (dismissed):

PR #827, opened 2023-02-20
[1pt] PR: Fix for missing branch_id.lst file

RobHanna-NOAA:
FIM 4.2.0.0. was throwing errors for 14 HUCs that did not have any level paths. These are HUCs that have only stream orders 1 and 2 and are covered under branch zero, but no stream orders 3+ (no level paths).  This has now been changed to not throw an error but continue to process of the HUC.

### Changes

- `src`
    - `run_unit_wb.sh`: Test if branch_id.lst exists, which legitimately might not. Also a bit of text cleanup.

## Testing

Ran fim_process_unit_wb.sh with a valid HUC and one invalid HUC to ensure the fix is correct. An invalid HUC is 01010011. Prior to the fix, the unit_errors folder would have a HUC unit log in the folder, but will not exists post fix. Pre-fix will show only partial cleanup of the branch_zero (branches/0) folder and you will specifically  notice the dem_meters_0.tif file will exist. Post fix, the dem_meters_0_tif will not exist as the second branch zero cleanup will have completed.

A few other invalid HUC's are 04160003 and 12110203.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,88 +1,14 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
-## v4.1.3.0 - 2023-02-13 - [PR#812](https://github.com/NOAA-OWP/inundation-mapping/pull/812)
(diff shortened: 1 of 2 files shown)


robgpita-noaa (approved):
Looks straightforward enough.

CarsonPruitt-NOAA (approved):

PR #835, opened 2023-03-02
[5 pt] PR: Fix for CatFIM sites sharing same LID and workaround for WRDS slowdown

LauraKeys-NOAA:
A bug was reported in the CatFIM flow-based service that showed different flood polygons sharing the same NWS LID incorrectly. This PR provides a fix for this bug, which was caused by not resetting a list of tifs to format, for each LID as maps and their metadata were being created. This merge also creates a workaround solution for a slowdown that was observed in the WRDS location API where feature_id metadata was taking over 2 seconds to be returned per request, causing Flow-Based and possibly Stage-Based to balloon in run time (multiple days).

## Changes

- `tools/generate_categorical_fim_mapping.py` resets the list of tifs to format for each LID within the loop that does the map processing, instead of only once before the start of the loop
- `tools/tools_shared_functions.py`:
  - adds a try-except block around code that attempted to iterate on an empty list when the API didn't return relevant metadata for a given feature ID (this is commented out, but may be used in the future once WRDS slowdown is addressed)
  - Uses a passed NWM flows geodataframe to determine stream order
- `/tools/generate_categorical_fim_flows.py': Adds multiprocessing to flows generation and uses `nwm_flows.gpkg` instead of the WRDS API to determine stream order of NWM feature_ids
- `/tools/generate_categorical_fim.py`: Refactor to allow for new NWM filtering scheme. Bug fix in multiprocessing calls for interval map production.

## Testing

1. Tested the offending NWS LID VCAC1 and one of the 3 sites that was being incorrectly labeled, SRFC1 (problem existed)
2. Tested SRFC1 on its own (problem went away)
3. Realized cause of problem, fixed problem, and then tested VCAC1 and SRFC1 again (problem fixed)
4. Tested VCAC1 with SRFC1, SESC1, and SCPC1 (problem fixed)
5. Ran Flow-Based and Stage-Based CatFIM and confirmed that run time had shortened (around 9-12 hours)


BradfordBates-NOAA (approved):
I ran SRFC1and VCAC1 and confirmed that they LIDs are unique between the two layers.

RobHanna-NOAA (dismissed):
Full successful tests against the 4.3.0.0 outputs was completed for both flow based and stage based.

PR #841, opened 2023-03-07
[1pt] PR: Delete intermediate agreedem files

### LINKED ISSUE #839: [5pt] Agree files not being cleaned up in branch zero
Branch zero has three files starting with the name "agree_" and they likely should not be there are end of processing the branch. 

Total output size for a run should be appx 1.3 TB, but these agree files pushed that to 2.98 TBs.

Research when/how we make agree files and if we can remove them at the first branch zero cleanup or the second. First cleanup strongly preferred because of file sizes.

May require only updates to the deny list files. Also research if we need the agree files named already in the deny files. Maybe they were renamed.

We also need to go back and cleanup our dev enviro prev fim, plus s3 fim.

mluck:
Deletes intermediate files generated by `src/agreedem.py` by adding them to `config/deny_*.lst`. Closes #839.

## Additions

- `config/`
    - `deny_branch_zero.lst`, `deny_branches.lst`, `deny_branch_unittests.lst`: Added `agree_binary_bufgrid.tif`, `agree_bufgrid_zerod.tif`, and `agree_smogrid_zerod.tif`
    - `deny_unit.lst`: Added `agree_binary_bufgrid.tif`, `agree_bufgrid.tif`, `agree_bufgrid_allo.tif`, `agree_bufgrid_dist.tif`,  `agree_bufgrid_zerod.tif`, `agree_smogrid.tif`, `agree_smogrid_allo.tif`, `agree_smogrid_dist.tif`, `agree_smogrid_zerod.tif`

### DIFF
--- a/config/deny_branch_unittests.lst
+++ b/config/deny_branch_unittests.lst
@@ -1,12 +1,15 @@
 # List of files for branches to delete
 # Use comment to allow list the file
 # Use curly braces to denote branch_id
+agree_binary_bufgrid.tif
 agree_bufgrid.tif
 agree_bufgrid_allo.tif
 agree_bufgrid_dist.tif
+agree_bufgrid_zerod.tif
 agree_smogrid.tif
 agree_smogrid_allo.tif
 agree_smogrid_dist.tif
+agree_smogrid_zerod.tif
 catch_list_{}.txt
 coordFile_{}.txt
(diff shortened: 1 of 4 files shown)


robgpita-noaa (approved):
Looks good to me, I did a test run, and none of those files you added to the deny lists were kept.

PR #845, opened 2023-03-15
[1pt] PR: Revise masking levee-protected areas from inundation

### LINKED ISSUE #822: [13pt] Levees are masking too many areas
The stream-order-based method for masking levees masks too many branches. I suggest using the following logic to join levee protected areas to their levee polylines and then using the polylines to determine which branches should get masked. Obviously, this is just a suggestion -- feel free to change it up if you have a different workflow that might work better. 

1. Acquire all levee polylines. Currently, we filter out the levees that do not have elevation data tied to them.
    - Change the levee burning module to filter the levees without elevation data since we're no longer doing it in the acquisition step.
2. For each HUC, clip both the levee polylines and levee protected areas to the WBD boundary.
3. For each levee protected area (polygon) within each HUC:
    - Use the `systemId` field to find which levee lines are associated with the selected area.
    - If a protected area has no associated lines, do not mask any branches (this levee protected polygon may be protected against a river in an adjacent HUC, not the current one).
    - For each associated levee line, find the nearest level path (`nwm_subset_streams_levelPaths_dissolved.gpkg`)
    - Levees sometimes protect against more than one level path, so many levee protected areas should have more than one level path assigned for masking.
    - If ANY levee lines intersect with a level path, that level path should be removed from the list of masked branches (this is for branches that flow behind levees - see HUC `08020401` in second screenshot for an example).

HUC `08020302` - Levee in adjacent HUC is incorrectly masking
image

HUC `08020401` Branch `1533000005` - This is correct inundation within a protected area that we would like to keep in the upgraded algorithm.
image

This is a follow-up to the previous levee masking issue #759.

mluck:
This PR revises the methodology for masking levee-protected areas from inundation. It accomplishes two major tasks: (1) updates the procedure for acquiring and preprocessing the levee data to be burned into the DEM and (2) revises the way levee-protected areas are masked from branches.

(1) There are now going to be two different levee vector line files in each HUC. One (`nld_subset_levees_burned.gpkg`) for the levee elevation burning and one (`nld_subset_levees.gpkg`) for the levee-level-path assignment and masking workflow.

(2) Levee-protected areas are masked from inundation based on a few methods:
- Branch 0: All levee-protected areas are masked
- Other branches: Levee-protected areas are masked from the DEMs of branches for level path(s) that the levee is protecting against by using single-sided buffers alongside each side of the levee to determine which side the levee is protecting against (the side opposite the associated levee-protected area).

Closes #822.

### Additions
- `.gitignore`: Adds `.private` folder to keep unversioned code
- `data/`
    - `esri.py`: Class for querying and downloading ESRI feature services.
    - `nld/`
        - `levee_download.py`: Module that handles downloading and preprocessing levee lines and protected areas from the National Levee Database.
- `src/associate_levelpaths_with_levees.py`: Associates level paths with levees using single-sided levee buffers and writes to CSV to be used by `src/mask_dem.py`

### Changes
- `.config/`
    - `deny_branch_zero.lst`: Adds `dem_meters_{}.tif`
    - `deny_branches.lst`: Adds `levee_levelpaths.csv` and removes `nld_subset_levees_{}.tif`
    - `deny_unit.lst`: Adds `dem_meters.tif`
    - `params_template.env`: Adds `levee_buffer` parameter for levee buffer size/distance in meters and `levee_id_attribute`
- `src/`
    - `bash_variables.env`: Updates `input_nld_levee_protected_areas` and adds `input_NLD` (moved from `run_unit_wb.sh`) and `input_levees_preprocessed` environment variables
    - `clip_vectors_to_wbd.py`: Added the new levee lines for the levee-level-path assignment and masking workflow.
    - `delineate_hydros_and_produce_HAND.sh`: Updates input arguments
    - `mask_dem.py`: Updates to use `levee_levelpaths.csv` (output from `associate_levelpaths_with_levees.py`) to mask branch DEMs
    - `run_by_branch.sh`: Clips `dem_meters.tif` to use for branches instead of `dem_meters_0.tif` since branch 0 is already masked.
    - `run_unit_wb.sh`: Added inputs to `clip_vectors_to_wbd.py`. Added `associate_levelpaths_with_levees.py`. Processes `dem_meters.tif` and then makes a copy for branch 0. Moved `deny_unit.lst` cleanup to after branch processing.
    - `burn_in_levees.py`: Removed the unit conversion from feet to meters because it's now being done in `levee_download.py`.

### Removals
- `data/nld/preprocess_levee_protected_areas.py`: Deprecated

### Testing
Tested on a number of HUCs including 03140202, 08080202, 08080301, 08080302, and 08080401.

### Notes
- Usage for the new levee download workflow is simply: `python /data/nld/levee_download.py`. There will be 3 outputs written to `/data/inputs/nld_vectors/`. Whenever levees are updated, `/src/bash_variables.env` should also be updated to reflect the new filenames since the dates will be updated.
    - `System_Routes_NLDFS_5070_{YYMMDD}.gpkg`
    - `3d_nld_preprocessed_{YYMMDD}.gpkg`
    - `Leveed_Areas_NLDFS_5070_{YYMMDD}.gpkg`


mluck:
## Inundation results

### 08020302
In contrast to the FIM4/FIM3 comparison in #822: image

there is now inundation in the levee-protected area (gray) with the 100 year recurrence flow. 
image

However, notice in the image above that the inundation doesn't follow the mainstem (green line). This is not related to the levee-masking process but instead is caused by diverging NWM (green line) and DEM-derived (black dotted line) stream networks which cause the REM (brown shading) for the mainstem branch to stop before it enters the levee-protected area. So none of the inundation in the leveed area is derived from the mainstem branch but all of it is coming from the mainstem tributaries (since branch 0 is also masked).
image

### 08020401
Inundation behind the levee-protected area is retained as desired.
image

RobHanna-NOAA (dismissed):
Did alpha domain test, alpha test and eval plots adn looks good. Ryan also reviewed the science side.

RobHanna-NOAA (approved):

RyanSpies-NOAA (commented):
Successfully tested the NLD data acquisition.
Successfully tested fim_pipeline.sh using huc 08040205

Issue #985, opened 2023-09-05
[21pt] Remove GMS branch outlet backpools


CarsonPruitt-NOAA:
When a GMS branch flows into a larger river at its outlet, we allow it to inundate the upstream channel to show a sort of backwater effect. FIM users have expressed that the inundation in a river that doesn't have a NWM (or RFC) flow forecast above high water is confusing. Therefore, we should try to remove this effect.

The cause of this issue is the large outlet catchment that includes contributing area from the mainstem river. The screenshots below are in HUC `13080002` branch `6077000088`.

![Image](https://github.com/NOAA-OWP/inundation-mapping/assets/90792257/245a5043-2e63-4b64-b63c-60bde499d193)

![Image](https://github.com/NOAA-OWP/inundation-mapping/assets/90792257/d2ec58b8-97c8-403e-be0e-646af0cf0cb5)

CarsonPruitt-NOAA:
https://vlab.noaa.gov/redmine/issues/119807

CarsonPruitt-NOAA:
My idea is to use the pixel catchments (`gw_catchments_pixels_6077000088.tif`) and identify the largest pixel catchment that occurs at the end of the DEM-derived streamline.

![image](https://github.com/NOAA-OWP/inundation-mapping/assets/90792257/630e5b40-3843-4848-add7-71cfcc5c6d50)

CarsonPruitt-NOAA:
The PR where we trimmed the downstream end of the GMS branches is #614. This issue is somewhat related in that we want to prevent branches from inundating the mainstem rivers, but the solution will need to be different.

EmilyDeardorff:
### Notes from tagup with Carson:

**Issue paraphrase:**

It looks like an large, erroneous catchment is being drawn at the outlet of the GMS branch. This large catchment is causing an artificially large amount of water to get routed to the smaller stream instead of the main stem, which means more of the water that could be causing flooding is coming from the small stream instead of going directly into the larger river.

The root of the issue is more in the deriving of catchments rather than the actual backwater calculations. 

The challenge here will be to figure out how to automatically choose the correct area to exclude from the HydroID catchment (i.e. the green highlighted area in the sketch).

**Potential approaches:**

- delineate and physically delete the extra catchment area, OR
- just shorten the DEM-derived streamline so the last catchment won’t include that area
    → preferred

**Related issues:**

[...]

PR #1006, opened 2023-10-11
[1pt] PR: Remove GMS branch outlet backpools

EmilyDeardorff:
In some HUCs, an overly-large catchment appears at the outlet of the branch (as in issue #985). This catchment causes an artificially large amount of water to get routed to the smaller stream instead of the main stem, which means more of the water that could be causing flooding is coming from the small stream instead of going directly into the larger river. This issue can be mitigated by trimming the levelpath just above the outlet and removing the offending pixel catchment from the pixel catchments and catchment reaches files. 

The branch outlet backpool issue is identified based on two criteria: 
  1. There is a pixel catchment that is abnormally large (more than two standard deviations above the mean.)
  2. The abnormally-large pixel catchment occurs at the outlet of the levelpath.
  
If both criteria are met for a branch, then the issue is mitigated by trimming the flowline to the third-to-last point.  The script also reads in the pixel catchments raster and the reaches raster and removes the inforrectly large outlet pixel catchment. This mitigates the branch outlet backpool error.

### Additions

- `src/mitigate_branch_outlet_backpool.py`
  - Detecting the branch outlet backpool error:
    - Added the catchment pixels file as an input to the main function
    - Added the `catch_catchment_size_outliers()` function to catch catchment size outliers in the catchment pixels layer
    - Added the `get_raster_value()` function to extract the catchment ID from the catchment_pixels geometry for a point feature (branch outlet backpool criteria 1)
    - Added the `check_if_ID_is_outlet()` function to test whether any of the outlier catchments are occurring at the levelpath outlet (branch outlet backpool criteria 2)

  - Mitigating the branch outlet backpool error:
    - If both branch outlet backpool criteria are met, the snapped point is set to be the penultimate vertex and then the flowline is trimmed to that point (instead of the last point). 
    - Trim the `gw_catchments_pixels_<id>.tif` and `gw_catchments_reaches_<id>.tif` rasters by using `gdal_polygonize.py` to polygonize the `gw_pixel_catchments_<id>.tif` file, creating a mask that excludes the problematic pixel catchment, and then using that mask to trim the pixel catchment and catchment reaches rasters.

### Changes

- `src/delineate_hydros_and_produce_HAND.sh`
  - Add the `mitigate_branch_outlet_backpool.py` module to run after the  `Gage Watershed for Pixels` step. 

- `src/split_flows.py`
  - Documentation and readability improvements: 
    - Improved documentation of script inputs at beginning of script
    - Added comments to better describe processing steps
    - Added headers to code sections

### Testing
- Tested updates on HUC 13080002.

_You may update this checklist before and/or after creating the PR. If you're unsure about any of them, please ask, we're here to help! These items are what we are going to look for before merging your code._

- [x] Passes all unit tests locally (inside interactive Docker container, at `/foss_fim/`, run: `pytest unit_tests/`)
- [x] [Reviewers requested](https://help.github.com/articles/requesting-a-pull-request-review/)


CarsonPruitt-NOAA:
I received this error when testing:
```
-----------------------------------------------------------------
Split Derived Reaches 13080002 6073000086
Thu Oct 12 20:30:52 UTC 2023
Loading data ...
Trimming DEM stream to NWM branch terminus...
Traceback (most recent call last):
  File "/foss_fim/src/split_flows.py", line 574, in <module>
    split_flows(**args)
  File "/foss_fim/src/utils/shared_functions.py", line 193, in wrapper
    func(*args, **kwargs)
  File "/foss_fim/src/split_flows.py", line 290, in split_flows
    if outlet_flag == True:
UnboundLocalError: local variable 'outlet_flag' referenced before assignment
```

EmilyDeardorff:
An unforseen issue is that `gw_catchments_reaches_$current_branch_id.tif`, which is an input to the new processing presented in this PR, actually is created in the step AFTER `split_flows.py` is run in `delinate_hydros_and_produce_HAND.sh` (in the section titled `GAGE WATERSHED FOR REACHES`, which runs `gagewatershed` from tauDEM). 

In order to move forward, we have to decide whether between a few options: 
1. Adjust the processing order so the catchment_pixels file is available for split_flows.py processing 
     _-> Probably wouldn't make sense because other outputs of the split_flows.py script are used as input_ 
2. Use a different catchment/catchment pixels file to find outlier catchments 
     _-> Could work, but only if a suitable replacement file is available_
3. Add the backpool error detection and mitigation steps to occur AFTER the  `GAGE WATERSHED FOR REACHES` section 
     _-> We would just want to be careful that we don't overwrite anything added to those files. One option would be to add this functionality into split_flows.py, another option would be to put it in a different python file entirely._

CarsonPruitt-NOAA:
I'm thinking that option 2 might be the best here. The more I think about our current strategy (outlier pixel catchments), the more I think that we could be introducing errors in other places. I'll make a tagup for us on Monday to discuss.

src/mitigate_branch_outlet_backpool.py
```diff
@@ -0,0 +1,286 @@
...
+from tqdm import tqdm
+from collections import Counter
+
+
+import build_stream_traversal
+from utils.fim_enums import FIM_exit_codes
+from utils.shared_functions import getDriver, mem_profile
+from utils.shared_variables import FIM_ID
+
+@mem_profile
```
robgpita-noaa, line 25: @EmilyDeardorff This just caught my eye. The `@mem_profile` decorator function has been removed in the latest dev branch. You would probably catch this after merging the latest dev, but figured I'd give you a heads up.

EmilyDeardorff: Fixed this! Good catch,

CarsonPruitt-NOAA:
I received this error when running HUC `12040101`:
```
-----------------------------------------------------------------
Catching and mitigating branch outlet backpool issue 12040101 1440000016
Thu Jan 25 21:30:15 UTC 2024

Non-branch zero, loading data for test ...
Catchment geom found, testing for backpool criteria...
20 outlier catchment(s) found in catchment size.
Flagged catchment(s) detected. Testing for second criteria.
Incorrectly-large outlet pixel catchment detected. Snapping line points to penultimate vertex.
Traceback (most recent call last):
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 464, in <module>
    mitigate_branch_outlet_backpool(**args)
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 308, in mitigate_branch_outlet_backpool
    thirdtolast_point = split_flows_last_geom['geometry'].apply(extract_3rdtolast_point).apply(Point)
  File "/usr/local/lib/python3.8/dist-packages/geopandas/geoseries.py", line 645, in apply
    result = super().apply(func, convert_dtype=convert_dtype, args=args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/pandas/core/series.py", line 4630, in apply
    return SeriesApply(self, func, convert_dtype, args, kwargs).apply()
  File "/usr/local/lib/python3.8/dist-packages/pandas/core/apply.py", line 1025, in apply
    return self.apply_standard()
  File "/usr/local/lib/python3.8/dist-packages/pandas/core/apply.py", line 1076, in apply_standard
    mapped = lib.map_infer(
  File "pandas/_libs/lib.pyx", line 2834, in pandas._libs.lib.map_infer
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 119, in extract_3rdtolast_point
    return line.coords[-3]
  File "/usr/local/lib/python3.8/dist-packages/shapely/coords.py", line 34, in __getitem__
    raise IndexError("index out of range")
IndexError: index out of range
Command exited with non-zero status 1
```

CarsonPruitt-NOAA:
Here's a way that I was looking at visualizing these backpool catchments. Each branch has it's removed backpool catchment area as a color-coded polygon. I.e., the dotted-orange polygon represents the backpool catchment area that was removed from the orange-colored branch.

Based off what I've seen so far, this is working exactly how we intended! 

Untitled

CarsonPruitt-NOAA:
As expected, there is a difference in the synthetic rating curves. However I didn't expect there to be this large of a difference. The map on the left shows the previous extent of the last catchment in a dotted outline (same color as map in previous comment). The SRC comparison chart on the right shows the next upstream SRC to show that the new SRC more closely matches others on this branch.
Untitled

CarsonPruitt-NOAA:
I was able to find this branch `1469000015` in HUC `12040103` that should probably be trimmed, but is not being picked up by your algorithm. This branch does have some other fairly large catchments in it, so my hunch is that they are causing the standard deviation for pixel catchment size to be abnormally high in this branch and that's why it's not being identified. Can you look into this one?

image

EmilyDeardorff:
Rob Hanna found this error during his tests on Feb 14th:

```-----------------------------------------------------------------
Catching and mitigating branch outlet backpool issue 01070006 6742000046
Fri Feb  9 18:44:43 UTC 2024

Non-branch zero, loading data for test ...
Catchment geom found, testing for backpool criteria...
24 outlier catchment(s) found in catchment size.
Flagged catchment(s) detected. Testing for second criteria.
Incorrectly-large outlet pixel catchment detected.                       Snapping line points to penultimate vertex.
/usr/local/lib/python3.10/dist-packages/geopandas/geodataframe.py:1443: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  super().__setitem__(key, value)
Extract third-to-last point of last geometry.
/usr/local/lib/python3.10/dist-packages/shapely/measurement.py:74: RuntimeWarning: invalid value encountered in distance
  return lib.distance(a, b, **kwargs)
Traceback (most recent call last):
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 537, in <module>
    mitigate_branch_outlet_backpool(**args)
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 375, in mitigate_branch_outlet_backpool
    trimmed_flows, inital_length_km = snap_and_trim_splitflow(pt_3tl_geom, split_flows_geom)
  File "/foss_fim/src/mitigate_branch_outlet_backpool.py", line 150, in snap_and_trim_splitflow
    if flow.index != outlet_point.index:
  File "/usr/local/lib/python3.10/dist-packages/pandas/core/ops/common.py", line 81, in new_method
    return method(self, other)
  File "/usr/local/lib/python3.10/dist-packages/pandas/core/arraylike.py", line 44, in __ne__
    return self._cmp_method(other, operator.ne)
  File "/usr/local/lib/python3.10/dist-packages/pandas/core/indexes/base.py", line 6758, in _cmp_method
    raise ValueError("Lengths must match to compare")
ValueError: Lengths must match to compare
Command exited with non-zero status 1
	Command being timed: "/foss_fim/src/run_by_branch.sh 01070006 6742000046"
	User time (seconds): 64.78
	System time (seconds): 47.21
	Percent of CPU this job got: 161%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:09.52
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 457400
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 1527592
	Voluntary context switches: 11121
	Involuntary context switches: 6348276
	Swaps: 0
	File system inputs: 0
	File system outputs: 770808
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 1```

EmilyDeardorff:
**Multiple NextDownID = -1 Issue**

I traced the branch error that Rob pointed out in his Feb 14th tests back to the fact that the branch in question had multiple branches where the `NextDownID` column had a value of -1. This is an issue because the `NextDownID` column is used to determine which stream segment contains the branch outlet. This is an uncommon situation but can occur in coastal branches or in branches with waterbodies.   

I've added a workaround that skips backpool mitigation on the branch if this issue is detected. I will add this information to a GitHub ticket so that a more comprehensive workaround can be added as an enhancement in the future.

EmilyDeardorff:
**Comparison Map**

I've run our current dev version and the dev-branch-outlet-backpools branch on HUC 13080002, which is the original HUC where this issue was pinpointed. Below is a visual comparison of inundation in branch 6077000088 from dev (on the left) compared to the dev-branch-outlet-backpools branch. This figure shows that the branch outlet backpools module mitigates the backpooling effect observed on the left. The middle panel shows the section of the branch catchments that was was causing the backpooling issue and was therefore removed (depicted by cross-hatching). 

Backpools Before and After

CarsonPruitt-NOAA (approved):

CarsonPruitt-NOAA (approved):

CarsonPruitt-NOAA (approved):

mluck:
I'm wondering why there's a difference in inundation between the left and right images below. Assuming the discharge is the same, are these from different FIM versions or is the backpool algorithm doing something here?
image

CarsonPruitt-NOAA:
There absolutely is a difference in the inundation that you can see in that map. However, I would expect it to go the opposite way - for there to be more inundation after the backpool fix due the change in SRC [seen here](https://github.com/NOAA-OWP/inundation-mapping/pull/1006#issuecomment-1915554313). This is a good catch, Matt. It's still curious because we tested this FB before merging and there wasn't any difference in skill metrics.

@EmilyDeardorff For both of these maps, are you using the standard, composited branch zero/GMS inundation?

mluck:
Since we were trying to find the source of the errors between 4.4.9.0 and 4.4.13.1, I ran two BLE HUCs (12040101 and 12090301) and it looks like this FB did change the CSI in both of the HUCs I looked at (red = decrease, green = increase). The number of branch errors (exit status 61) in two HUCs (09010004 and 12100303) also seem to have increased in this FB.
image

mluck:
The blue areas (a-c) are areas of inundation that were removed by this FB, while the orange (d) was added. I thought that this FB only works at the end of levelpaths but it appears that this is also removing area along levelpaths. It looks like maybe this is because it's using the undissolved levelpaths?
image

CarsonPruitt-NOAA:
We're using `NextDownID == '-1'` to find the terminal segment of each level path ([See here](https://github.com/NOAA-OWP/inundation-mapping/blob/8e88f1800642c421390b061bdd9e15294208a051/src/mitigate_branch_outlet_backpool.py#L298)). Maybe the segments you identified have a `NextDownID` of -1?

mluck:
I took a closer look at the four places I pointed out earlier. It seems that this PR is doing what it should with regards to inundation (and this is likely affecting the metrics but not sure about the increase in branch errors). Here are some detailed inundation images (left = before, right = after this PR). The red line in the first two images below are DEM-derived streams with `NextDownID == -1`.

My only remaining question is why in image (a) and the Laredo example (that I pointed out above) are the differences in inundation above the last reach, not in the last reach? At first I thought it was because the points might be digitized in the opposite direction (downstream to upstream) but it seems like that's not the case since the large catchment area on the mainstems are being removed correctly, while it seems that the change in inundation at (d) happens at the last reach.

(a) 
image

(b), (d) 
image

(c)
image

mluck:
In the example (a) above, the SRC for the last (outlet) reach is changed slightly but the SRC of the penultimate reach is very different and would seem to account for the decreased inundation in 25070004 (lighter shaded blue). The catchments are the same before and after this FB.
image

SRC for the last (outlet) reach (25070005)
image

SRC for next reach upstream (25070004)
image

mluck:
This is the effect on the agreement raster. Interesting that the next-to-last reach was initially over-inundated.

Before (4.4.12.0)
image

After (4.4.13.0)
image

mluck:
All of the Manning's parameters are identical between the two versions. The only thing that is changing in the next to last reach is the spatial calibration coefficient. Here's the comparison of the highest stage in 4.4.12.0 (top) and 4.4.13.0 (bottom):
image

CarsonPruitt-NOAA:
@RyanSpies-NOAA Can you take a look at why the spatial calibration coefficient is so different in the newest version? This is HUC `12090301` LID `brtt2` and I believe the feature_ID for the catchment in question is `5789218`.

PR #1554, opened 2025-06-12
[3pt] PR: Lofi performance tweaks

groutr:
Some minor tweaks that should help performance of lofi.

Performance enhancements:
* Vectorized `generate_streamflow_percentiles`
* Removed unnecessary copies
* Reduced unnecessary iterations over collections

Bug fixes:
* Properly compare floats
* Remove threading when writing dataset
* Exit non-zero when exception is raised
* Cleanup defaults in cli

User facing:
* Create compressed and tiled GeoTiff rasters

ping @GregoryPetrochenkov-NOAA

### DIFF
--- a/tools/lofi/probabilistic_inundation.py
+++ b/tools/lofi/probabilistic_inundation.py
@@ -665,18 +665,14 @@ def inundate_probabilistic(
         ds.close()
 
     if output_vector is True:
+        def _make_geometry(shapes):
+            for p, v in shapes:
+                yield shape(p), v
+
         with rasterio.open(out_rast, 'r') as rst:
             shapes = rasterio.features.shapes(rst.read(1), mask=None, transform=rst.transform)
-
-            polygons = []
-            for geom, value in shapes:
-                polygon = shape(geom)
-                polygons.append((polygon, value))
-
-            data = []
-            for polygon, value in polygons:
-                data.append({'geometry': polygon, 'value': value})
-            gdf = gpd.GeoDataFrame(data, crs=raster_crs)
+            gdf = gpd.GeoDataFrame(_make_geometry(shapes), columns=['geometry', 'value'], crs=raster_crs)
+            gdf = gdf.set_geometry('geometry')
             gdf.to_file(os.path.join(base_output_path, output_file_name))
 
     for file in percentile_files:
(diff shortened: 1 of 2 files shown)


RobHanna-NOAA:
@GregoryPetrochenkov-NOAA : Can you help update this PR for Ryan for conventions such as PR titles, changelog, projects, etc? :)
And maybe you can do linting for him too if that is ok.  @groutr :  I am happy to show these to you if you like, or we can take care of it for you as you are not used to our team conventions.

groutr:
I can probably take care of the linting. As far as the other stuff, idk what your conventions are, but this code should still be tested.

groutr:
I removed threading support from raster output block of code. GDAL (and by extension rasterio) do not support sharing a dataset among threads. Each thread must create its own dataset handle. Reading/writing to the same dataset from multiple threads though sounds like a race condition waiting to happen.

I also enabled tiling and compression on the output TIF file.
ping @RobHanna-NOAA @GregoryPetrochenkov-NOAA

tools/lofi/probabilistic_inundation.py
```diff
@@ -658,62 +611,32 @@ def inundate_probabilistic(
...
-
-    def _vprint(message, verbose):
-        if verbose:
-            print(message)
-
     out_rast = os.path.join(base_output_path, output_file_name.replace(".gpkg", ".tif"))
-    with rasterio.open(out_rast, "w+", **profile) as write_rst:
-        dgen = __data_generator(datasets, list(percentiles.keys()), windows, write_rst)
-        results = {executor.submit(merge_percentiles, *wg): 1 for wg in dgen}
+    with rasterio.open(out_rast, "w+", **profile, compress='DEFLATE', tiled=True) as write_rst:
```
GregoryPetrochenkov-NOAA: The compress and tiled arguments existed in profile already so it throws `TypeError: rasterio.open() got multiple values for keyword argument`.  Could you please instead add this to profile update on line 612 as follows:  `profile.update(dtype=np.int8, compress='DEFLATE', tiled=True)`.

GregoryPetrochenkov-NOAA (changes requested):
I did some local testing and this looks good minus one change that needs to be addressed.  After this change and a test in the step functions we should be set to approve and merge.

tools/lofi/probabilistic_inundation.py
groutr: Added. It now only falls back to DEFLATE if no compression exists.

PR #1502, opened 2025-04-24
[5pt] PR: Implement general-purpose multiprocessing runner with logging

AliForghani-NOAA:
This PR introduces a multiprocessing utility that can be reused across different workflows and projects.

### Key features:
- Parallel Task Execution: Uses `ProcessPoolExecutor` to run a set of independent tasks in parallel, with configurable number of workers.
- Robust Logging:
  - File-based logging using a centralized `file_logger` for recording of individual task results and errors.
  - Screen output via a multiprocessing-safe `screen_queue`, printed by a dedicated background thread without interrupting the main progress bar.
  - Task functions always receive three additional arguments of `file_logger`, `screen_queue`, and `task_id` that allow logging for each individual task.
- Progress Monitoring: Displays real-time progress using `tqdm`, with success or failure messages for each task.

#### 1. Prepare the list of task arguments

Create a list of dictionaries, where each dictionary contains the necessary inputs for a single task.  Also, it is helpful to sort the tasks by their identifier, as this provides a clearer sense of progress when reviewing the logs during the run.
For example, to process each HUC's geometry and associated catchment boundaries:

```python
import os
import numpy as np
import geopandas as gpd

output_dir = "outputs/roads/HUCs_roads_test/"

# Build the tasks argument list
huc_gpd = gpd.read_file("wbd/WBD_National_HUC8_EPSG_5070_HAND_domain.gpkg")
huc_numbers = sorted(np.unique(huc_gpd["HUC8"])) # Sort tasks to help track progress in logs

tasks_args_list = []
for HUC_no in huc_numbers:
    this_huc_info = huc_gpd[huc_gpd["HUC8"] == HUC_no]
    this_huc_geom = this_huc_info.iloc[0]["geometry"]

    tasks_args_list.append({
        "HUC_no": HUC_no,
        "huc_geom": this_huc_geom,
        "catchment_path": f"data/inputs/pre_clip_huc8/20250218/{HUC_no}/nwm_catchments_proj_subset.gpkg",
        "output_dir": output_dir
    })
```

#### 2. Define the main task function (**see the note below**)

This **function** manages the work per task and should handle any sub-functions internally. See the section `Updates Introduced After PRs #1660 and #1647` for more details.  

```python
import traceback

def task_fn(HUC_no, huc_geom, catchment_path, output_dir, file_logger, screen_queue, task_id):
    file_logger.info(f"Started processing {task_id}")
    screen_queue.put(f"Started processing {task_id}")
    try:
[...]

PR #1660, opened 2025-09-22
[1pt] PR: MP Runner bug fix

### LINKED ISSUE #1659: [5pt] Bug in updated MP runner
The updated multiprocessing (MP) runner merged last Friday through [PR 1647 ](https://github.com/NOAA-OWP/inundation-mapping/pull/1647) is causing a failure when running `data/roads/pull_osm_roads.py` for multiple HUCs. Command used:

```bash
python foss_fim/data/roads/pull_osm_roads.py \
  -p data/inputs/pre_clip_huc8/20250917/ \
  -o outputs/RobPR/run2 \
  -j 2 \
  -lh "21010008 10300103"
```

Below is the error message I get:

Image

Note: [PR 1647](https://github.com/NOAA-OWP/inundation-mapping/pull/1647) was merged before review was completed:

RobHanna-NOAA:
The file shared_functions.py had a bug in it related to a object length check. During that processes we discovered some enhancements to logging tools and run_with_mp.  The return code system was changed now where 1 = success, 0 = fail but do not abort and -1 means fail and abort run.  

This triggered changes to scripts that use the logging or run_with_mp tools for their return values from their multi-proc functions.  The change affected four data load scripts, and all were retested for correct data loading. A fifth data load script, pull_osm_bridges.py, was updated but only for a comment and was not affected by other changes in this PR.

There was also a change to shared_functions.setup_file_logger. It no longer takes a full path and file name for  a log file, but two separate variables, one for the log folder, the other for a log file prefix. A date will be added automatically to the log file name.

### Changes

- `data`
    - `bridges\make_dem_dif_for_bridges.py`:  Updated call to setup_file_logger plus changes related to run_with_mp. A bit of additional error handling was also added.
    - `bridges\pull_osm_bridges.py`:  Just a comment added
    - `nfhl\download_fema_nfhl.py`:  Updated call to setup_file_logger plus changes related to run_with_mp.
    - `roads\pull_osm_roads.py`:  Updated call to setup_file_logger plus changes related to run_with_mp.
- `src\shared_functions.py`: as described.
    - `usgs\get_usgs_rating_curve.py`:  Updated call to setup_file_logger plus changes related to run_with_mp. A bit of additional error handling was also added. A small fix for rounding was also fixed.
    
Closes: [1659](https://github.com/NOAA-OWP/inundation-mapping/issues/1659)

### Testing

This changes affects four scripts:  pull_osm_roads.py, make_dem_dif_for_bridges.py, download_fema_nfhl.py and get_usgs_rating_curves.py.  All were retested and all had a way to quickly stub test with small batches.

    - [ ] Yes
    - [x] No  (f no.. skip the rest of the Deployment Plan section)
    
       - [ ] Yes
       - [ ] No
       - [ ] No
       - [ ] Yes
           -  Require new pre-clip set or any other data reloads, such as DEMS, osm, etc. ie.. pre-requisite re-data upstream of your input  changes.
                - [ ] Yes
                - [ ] No
           -  Has the inputs been copied/exist in all five enviros:
                 - [ ] FIM EFS  
                 - [ ] FIM S3
                 - [ ] ESIP
                 - [ ] Dev1
                 - [ ] UCS2
            
    - [ ] Yes

- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments


src/utils/shared_functions.py
AliForghani-NOAA, line 1: [This if statement](https://github.com/NOAA-OWP/inundation-mapping/blob/be4eb0c87f778ad2b6b9900f08cbbab94dd2b2b6/src/utils/shared_functions.py#L334) is redundant.

AliForghani-NOAA, line 1: We may need to update the [setup_mp_file_logger](https://github.com/NOAA-OWP/inundation-mapping/blob/be4eb0c87f778ad2b6b9900f08cbbab94dd2b2b6/src/utils/shared_functions.py#L101) function so that each `logger_name` is uniquely bound to a single `log_file_path`, and each `log_file_path` is uniquely bound to a single `logger_name`. The function should raise an error if a `logger_name` is reused with a different file path, or if a file path is reused with a different `logger_name`. However, repeated calls with the same (logger_name, log_file_path) pair are permitted, allowing safe reuse of the same logger. The same rule may need to be applied for the new `setup_file_logger` function as well. 

```python

# Registry: name -> path
_LOGGER_REGISTRY = {}

def setup_mp_file_logger(log_file_path: str, logger_name: str, level=logging.DEBUG) -> logging.Logger:
    """
    Create a logger bound by a strict bijection:
      - Each logger_name maps to exactly one log_file_path.
      - Each log_file_path may only be used by one logger_name.
    """
    if not logger_name:
        raise ValueError("logger_name must be a non-empty string.")

    abs_path = os.path.abspath(log_file_path)
    os.makedirs(os.path.dirname(abs_path), exist_ok=True)

    # Check name -> path
    if logger_name in _LOGGER_REGISTRY and _LOGGER_REGISTRY[logger_name] != abs_path:
        raise ValueError(
            f"Logger '{logger_name}' already bound to '{_LOGGER_REGISTRY[logger_name]}', cannot bind to '{abs_path}'."
        )

    # Check path -> name (reverse lookup)
    for name, path in _LOGGER_REGISTRY.items():
        if path == abs_path and name != logger_name:
            raise ValueError(
                f"Path '{abs_path}' is already bound to logger '{name}', cannot assign to '{logger_name}'."
            )

    # Register if new
    _LOGGER_REGISTRY.setdefault(logger_name, abs_path)

    logger = logging.getLogger(logger_name)
    logger.setLevel(level)

    if not logger.handlers:
        fh = logging.FileHandler(abs_path)
        fh.setLevel(level)
        fh.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(message)s"))
        logger.addHandler(fh)
        logger.propagate = False

    return logger

```

src/utils/shared_functions.py
```diff
@@ -348,7 +348,7 @@ def log_worker(queue):
                         # IMPORTANT NOTE:
                         #    This extracts the first item only.
                         results[task_id] = rtn_value[0]
-                    if len(rtn_value > 1):
+                    if len(rtn_value) > 1:
```
AliForghani-NOAA: No need to restrict the list to a single item. Instead, allow task functions to return a list (along with the return code) that can hold any number of items. This removes the need to check the list length and makes the pattern simpler. For example:

```python
rtn_code, rtn_list = future.result()
results[task_id] = rtn_list
```

src/utils/shared_functions.py
AliForghani-NOAA, line 1: The current [status codes](https://github.com/NOAA-OWP/inundation-mapping/blob/be4eb0c87f778ad2b6b9900f08cbbab94dd2b2b6/src/utils/shared_functions.py#L200C8-L200C37) are counterintuitive and error-prone (easy to be confused):

```python
# Current (confusing):
0 = Success
1 = Critical  
2 = Failure
```

**Proposed improvement:**
```python
 1 = Success    # Positive = good 
 0 = Regular Failure    # Zero = neutral   
-1 = Critical   # Negative = abort 
```

**Benefits:**
- Intuitive: no need to memorize arbitrary mappings 
- Mathematical: easy checks with `if result > 0`, `elif result == 0`, `else`
- Extensible: can add more levels in future (if needed) without breaking the pattern

RobHanna-NOAA: fix coming in next checkin

src/utils/shared_functions.py
RobHanna-NOAA: This could result in a list inside a list.  ie) results[task_id] = {another list).  Do you think this is ok? I guess it wouldn't really be a big deal. Rare I would guess, but nothing saying we can't. Thoughts?

src/utils/shared_functions.py
RobHanna-NOAA: Love the idea. Will be in next checkin.

RobHanna-NOAA: Good idea on the setup_mp_file_logger. I am adding it in as is.  For the setup_file_logger, I am thinking there is a better way to do it anyways to avoid file log overwrites.  For that one, we don't let them give us a full log file name and path. instead we only let them give us a log folder path, then a log file prefix. We auto append a date/time.  
ie) 
def setup_file_logger(log_file_dir, log_file_name_prefix):

    file_dt_string = datetime.now(timezone.utc).strftime("%Y%m%d-%H%M")
    log_file_name = f"{log_file_name_prefix}_{file_dt_string}.log"
    log_file_path = os.path.join(log_file_dir, log_file_name)
    print(f"Logs saved to: {log_file_path}")

src/utils/shared_functions.py
AliForghani-NOAA: Thanks, the current method is correct allowing multiple items in the returned list.

src/utils/shared_functions.py
AliForghani-NOAA, line 1: I have two more suggestions:

1- The items in the lists returned by the try and except blocks must be consistent. If the try block returns [dataframe], the except block should return a placeholder such as [None]. Please add None [here ](https://github.com/NOAA-OWP/inundation-mapping/blob/7319ef63307875809c622e303aa74775700aee65/src/utils/shared_functions.py#L250)(and in the following line).

2- In [this ](https://github.com/NOAA-OWP/inundation-mapping/blob/7319ef63307875809c622e303aa74775700aee65/src/utils/shared_functions.py#L361)line, do we need to include warning as well as error? I think it is more clear that an error happens so the sub-task failed. If you agree, please remove warning.

RobHanna-NOAA: Good questions:
1) We never want anything returned like `[None]`, not even sure it is possible, it should be either `None` or `[]`.  We do handle both possibilities of those being returned.
      - (line 374/5:    # Some calling functions, may return a None, especially if it returned rtn_code == 0 (fail but continue)
                    if rtn_value is not None:
                        results[task_id] = rtn_value
      - however, I should update the notes on that where you pointed to. I will do that soon.

2 ) Humm... good question: I believe we do need something at least in the logs. We need to bring attention to the fact that either an error or warning occurred, which will roll up into the error log.  The user may not realize that some or all of his mp's are failing. But.. in theory they should have logged a message inside the child mp that uses the word "error" or "warning" so the rollup catches it.  My gut says leave it as is.  If the user starts seeing every single one of his/her child MP's failing.... they know they have a problem, instead of waiting hours later to see it in a log file.

RobHanna-NOAA: I just updated some inline notes where item number 1 which hopefully will help clarify the developers options. :)

AliForghani-NOAA: Rob,
1- Yes, it is both possible and desirable to have one or more `None` in the returned list (i.e., payload list) . This is important when tasks return mixed outputs like this:
success: [df1, df2, True],  
failure: [None, None, False]

The presence of those `None` values (in failure case) is essential in processing the returned results correctly. For example, if you want to have access to True and False items of the above lists.  

In addition, per the new rule below (as notes [here](https://github.com/NOAA-OWP/inundation-mapping/pull/1502) under `Updates Introduced After PRs 1660 and 1647`):

`Each task must return two items: return_code, payload_list`

This means that tasks can never return `None`. Both `try` and `except` blocks of the main task function MUST always return a list as their second retuned item. Therefore, [this check](https://github.com/NOAA-OWP/inundation-mapping/blob/0cae626c8bb8a25376084c12de9a6d443bc3c832/src/utils/shared_functions.py#L379C21-L379C46) is not needed--rtn_value is always a list.

 Of course, the payload list may be empty or contain `None` values as needed **to keep the number and order of items consistent between the success (`try`) and failure (`except`) cases**.

2- My only suggestion was to change

```python
tqdm.write(f"❌ Error or Warning reported for {task_id}.")
```

to

```python
tqdm.write(f"❌ Error reported for {task_id}.")
```

I’m not suggesting any changes related to the logs. I proposed this change because that message is only displayed when an error occurs (i.e., when the `except` block is triggered). Warnings do not trigger the except block, so including “warning” in this message could be misleading.
That said, if you prefer to keep the word *“warning”*, that’s fine.

RobHanna-NOAA: Huh.. you are right.. you can add an item to a list which is None.  I didn't know that.  I have added and checked in your recommendations. Huge thanks for helping me make this even more robust.

AliForghani-NOAA: Thanks Rob! Great improvements for the MP function.

AliForghani-NOAA (dismissed):

CarsonPruitt-NOAA (approved):

PR #1745, opened 2026-02-02
PR: [1pt] More efficient file read in FLASH FIM

groutr:
Restructured the reading of the grib file to make it more efficient. It reads the file once outside the loop instead of each loop iteration.

ping @CarsonPruitt-NOAA


groutr:
I realized the range processing loop doesn't depend on the file being open, so we can just read the band then close the file.

RobHanna-NOAA:
Rolled into PR [1803](https://github.com/NOAA-OWP/inundation-mapping/pull/1803):  Optimize FLASH FIM Workflow

Issue #1765, opened 2026-02-18
[8pt] No NWM streamlines in HUC


mluck:
In the v4.9.9.0 BED, two HUCs (04110002 and 04160003) exited with status 60 (example below).

```
Generating Level Paths for 04160003
Sorry, no branches exist and processing can not continue. This could be an empty file.
Command exited with non-zero status 60
	Command being timed: "/foss_fim/src/run_huc.sh"
	User time (seconds): 2.05
	System time (seconds): 0.62
	Percent of CPU this job got: 151%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.76
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 215752
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 29377
	Voluntary context switches: 750
	Involuntary context switches: 110
	Swaps: 0
	File system inputs: 112424
	File system outputs: 112224
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 60
```

mluck:
Looking back through the `inputs/pre_clip_huc8`, there have been no `nwm_subset_streams.gpkg` for these HUCs since 20250717. The original dataset does have streamlines for these HUCs including up to order 5, but none of the downstream ends of the streamlines intersect with the WBD boundary for that HUC.

mluck:
This the example from HUC 04110002 that includes parts of Cleveland and Akron, Ohio. 

<img width="746" height="619" alt="Image" src="https://github.com/user-attachments/assets/ed821be7-97f5-455b-9eea-00dc018a4395" />

There are two issues with the NWM stream line:
- it doesn't reach the HUC boundary (black line) at the edge of Lake Erie, which could be due to the WBD jutting into Lake Erie at the mouth of this river.
- the outlet `to` attribute of the final stream segment (red) should be 0.

So the terminal segment here isn't recognized as an outlet which causes the pre-clipping to fail and not generate any streams for this HUC.

<img width="1197" height="650" alt="Image" src="https://github.com/user-attachments/assets/7ea47514-caad-45be-8881-0572e5d6b491" />

PR #1783, opened 2026-03-12
[1pt] PR: Fix NWM stream outlet snapping

### LINKED ISSUE #1782: [5pt] Streams outside of the WBD buffer are snapped back to the WBD buffer
#1769 introduced a bug where NWM stream outlet segments that extend beyond the buffered WBD (green) are being snapped back to the buffered WBD (the red line), possibly causing erroneous flow and inundation. This may be occurring because the stream outlet line exits then re-enters and re-exits the buffered WBD.

Here's an example from HUC 05120111:
Image

mluck:
Resolves an issue causing stream outlet lines extending outside of the buffered WBD to be snapped back to the buffered WBD. 

NWM streams (`nwm_subset_streams.gpkg`) were updated in /data/inputs/pre_clip_huc8 with the date of 20260312 based on the previous preclip of 20260306.

Closes #1782.

### Changes

- `data/wbd/clip_vectors_to_wbd.py`: Ignores `linegeom` assignment if already assigned
- `src/bash_variables.env`: Updates preclip date to 20260312

### Testing
See https://github.com/NOAA-OWP/inundation-mapping/pull/1783#issuecomment-4048934197 below for testing results. Also tested on HUC 03100204 as a control.

    - [x] Yes
    - [ ] No  (f no.. skip the rest of the Deployment Plan section)
    
       - [ ] Yes
       - [x] No
       - [ ] No
       - [x] Yes
           -  Require new pre-clip set or any other data reloads, such as DEMS, osm, etc. ie.. pre-requisite re-data upstream of your input changes.
                - [x] Yes
                - [ ] No
           -  Has the inputs been copied/exist in all four enviros:
                 - [x] FIM EFS  
                 - [x] FIM S3
                 - [x] ESIP
                 - [x] Dev1
            
    - [x] Yes

- [x] Where applicable, has fim_pipeline been tested with multiple HUCs, including some other unaffected HUCs?

### Reviewer / Approver Checklist
- [ ] Where applicable, has fim_pipeline been tested with muliple HUCs, including some other unaffected HUCs?
- [ ] If there are new inputs, have you confirmed that they have been copied to all enviroments?


mluck:
I did pretty extensive testing on this to try to ensure that there no unintended effects. I ran a preclip on the UAT set of HUCs and analyzed the FIM pipeline and alpha test results based on the new preclip.

An example of the corrected streamline for HUC 04110002 (referenced in the original issue (https://github.com/NOAA-OWP/inundation-mapping/issues/1765#issuecomment-3947052718) is here, where the yellow lines are added to the end of the outlet stream to enforce correct streamflow:
image

For the UAT, the CSI metrics were nearly identical to v4.9.9.0 with the exception that the NWS action stage increased.
image

I looked at the individual BLE, NWS, and USGS and selected all of the ones that had changed negatively. In all of these cases, the issue was caused by a stream outlet not reaching the edge of the DEM (which is clipped to the buffered WBD) resulting in reverse flow. This PR resolved these issues, in most cases (but not all) increasing the CSI skill.

The site that had the greatest recession in skill was `ckln6`, especially for the lower magnitudes. 
image

For the action magnitude, this PR introduces some overinundation in the west (left) side of the evaluation domain. By the major magnitude, this inundation is incorporated into the true positive inundation and the skill increases. However, there was no inundation coming from branch 0 in the v4.9.9.0 version (right image)
image

Comparing branch 0 inundation across magnitudes shows that branch 0 inundation for v4.9.9.0 is spotty in the moderate and complete in the major magnitude.
image

At the downstream outlet, the dem_thalwegCond.tif in v4.9.9.0 has a channel still burned in while this feature branch doesn't.
image

At the upstream end, the issue seems to be coming from the fact that the NWM flowline crosses the HUC boundary as the red line in the left image flows downward (south) into the HUC, but it is connected to another network flowing upwards to the north. However, the NWM flow segment that was supposed to exit the buffered WBD didn't quite reach it, causing the reverse flow issue (top right). This was resolved by this PR (bottom right).
image

This had the effect of correcting the DEM-derived reaches (yellow) so that they didn't exit upwards at the HUC boundary to the north here but instead flowed to correct outlet in the lower left.
image

In another example, at `ekmi3` (HUC 04050001), the v4.9.9.0 outlet stream (yellow) stopped short of the buffered WBD. As it was extended by this PR (red segment). The values of `dem_burned` and the `dem_burned_filled - dem_burned`change drastically between the outlet pixel and the end of the v4.9.9.0 stream segment (yellow line), causing a reverse flow issue here.
image

Finally, for BLE HUC 12020007, there was a different issue. It appeared that there wasn't a reverse flow issue, but the dem_thalwegCond depths along the stream channel were different between v4.9.9.0 and this PR. The NWM streamlines were identical here, but the branch 0 DEM-derived reaches showed that the v4.9.9.0 had a reach (red line) that wasn't present in the PR version (yellow lines).
image

The difference in `dem_thalwegCond.tif` is on the left. This was traced to a reverse flow issue outside of the HUC, within the buffered WBD, that ran through what is an obvious issue with the DEM data. The stairstep pattern of coarser DEM cells is 1-2 meters higher than the neighboring DEM values (raw `dem_meters.tif` displayed here).
image

hhs732 (dismissed):
This PR was successfully tested on 6 HUC8s including `02030103;  05120201;  02050101;  03100204;  04110002;  04050001`. Eval plots were created for the test sample, and none showed an unexpected results.

hhs732 (dismissed):
Reviewed and approved.

CarsonPruitt-NOAA (approved):

PR #1805, opened 2026-04-08
[1pt] PR: Upgrade GDAL base image to v3.12.3 and resolve/upgrade Python dependencies

mluck:
Upgrades GDAL base image to v3.12.3 (ghcr.io/osgeo/gdal:ubuntu-small-3.12.3) and upgrades Python dependencies. There were a few major hurdles in upgrading beyond the previous GDAL v.3.8.4 primarily due to the fact that v3.8.4 was the last version to use Python 3.10 and GDAL v3.12.3 uses Python 3.12, including:
- Numerous Python dependency conflicts
- `numpy` versions >= 2.0.0 are incompatible with the GDAL v3.12.3 Docker image. `numpy` and `scipy` were forced to downgrade after the Python packages were installed via Pipfile.
- Python 3.12 manages packages in a virtual environment which made them unable to be found.
- A change in `gdal_rasterize` caused 0s to be written as nodata which tripped up `fim_pipeline.sh`.
- Incompatibility between TauDEM v5.3 and GDAL v3.12.3 resulted in an Error message. The outputs are still valid so the error message was suppressed.

In addition, an upgraded `pdal` was added to the Dockerfile and `pillow` was upgraded to v12.2.0.

The dockerfile.owp has been stripped way back to remove most code such as two stages, taudem and minimamba.  This results in us no longer being able to run most products such as fim-pipeline. The only scripts that still need to be run in the OWP servers are:
- CatFIM  (but no for much longer)
- get_usgs_rating_curves.py
- AWS CLI tools
- download_process_words.py
- eval_plots.py (but only with args to create fim_performance points and polys

Resolves #1735.

This PR also updates and closes 
#1833 - urllib3
#1819 - jupyter and jupyterlab
#1823 - jupyter-server
#1826 - mistune
#1812 - nbconvert
#1813 - python-dotenv
#1810 - pillow
and the requests package
as well as a wide number of other packages referenced only in the pipfile.lock

### Changes

- `Dockerfile.dev` and `Dockerfile.owp`: Upgrade GDAL base image to v3.12.3; adds system break flag for installing Python packages in an unmanaged environment; upgrades `pdal`; downgrades `numpy` and `scipy`
- `Pipfile` and `Pipfile.lock`: Upgrade and resolve Python dependencies and conflicts
- `.gitignore`: Updated to ensure the projects do not include juypter checkpoint files as well as local git repo files for when a user stores repo creates at the project level and not the global/server level.
- `src/`
[...]

PR #1862, opened 2026-06-15
[1pt] PR: Default to geoparquet vector outputs in FIM pipeline

### LINKED ISSUE #1853: [13pt] Geopackage SQLite errors
Errors such as `database is locked`, `disk I/O error`, `no such table`, and `failed to rollback` are being thrown sporadically while running `fim_pipeline.sh`. They are SQLite errors mostly related to writing Geopackages (.gpkg) in many of the Python and bash scripts. For example:

`src/filter_catchments_and_add_attributes.py`
```
Process catchments and model streams 03100204 0
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/geopandas/io/file.py", line 772, in _to_file_fiona
    colxn.writerecords(df.iterfeatures())
  File "/usr/local/lib/python3.12/dist-packages/fiona/collection.py", line 541, in writerecords
    self.session.writerecs(records, self)
  File "fiona/ogrext.pyx", line 1699, in fiona.ogrext.WritingSession.writerecs
fiona.errors.TransactionError: Failed to commit transaction

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/foss_fim/src/filter_catchments_and_add_attributes.py", line 133, in <module>
    filter_catchments_and_add_attributes(**args)
  File "/foss_fim/src/filter_catchments_and_add_attributes.py", line 98, in filter_catchments_and_add_attributes
    output_flows_filtered.to_file(
  File "/usr/local/lib/python3.12/dist-packages/geopandas/geodataframe.py", line 1632, in to_file
    _to_file(self, filename, driver, schema, index, **kwargs)
  File "/usr/local/lib/python3.12/dist-packages/geopandas/io/file.py", line 733, in _to_file
    _to_file_fiona(df, filename, driver, schema, crs, mode, metadata, **kwargs)
  File "/usr/local/lib/python3.12/dist-packages/geopandas/io/file.py", line 767, in _to_file_fiona
    with fiona.open(
  File "/usr/local/lib/python3.12/dist-packages/fiona/collection.py", line 682, in __exit__
    self.close()
  File "/usr/local/lib/python3.12/dist-packages/fiona/collection.py", line 659, in close
    self.flush()
  File "/usr/local/lib/python3.12/dist-packages/fiona/collection.py", line 649, in flush
    self.session.sync(self)
  File "fiona/ogrext.pyx", line 1708, in fiona.ogrext.WritingSession.sync
  File "fiona/ogrext.pyx", line 87, in fiona.ogrext.gdal_flush_cache
  File "fiona/_err.pyx", line 224, in fiona._err.GDALErrCtxManager.__exit__
fiona._err.CPLE_AppDefinedError: b'sqlite3_exec(COMMIT) failed: database is locked'
```

`run_by_branch.sh`
```
Subsetting vectors to branches 03100204 1691000007
Querying NWM streams ...
ERROR 1: sqlite3_exec(UPDATE gpkg_contents SET last_change = strftime('%Y-%m-%dT%H:%M:%fZ','now') WHERE lower(table_name) = lower('nwm_subset_streams_levelPaths_extended')) failed: database is locked
ERROR 1: sqlite3_exec(CREATE TRIGGER "trigger_insert_feature_count_nwm_subset_streams_levelPaths_extended" AFTER INSERT ON "nwm_subset_streams_levelPaths_extended" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count + 1 WHERE lower(table_name) = lower('nwm_subset_streams_levelPaths_extended'); END;) failed: database is locked
ERROR 1: sqlite3_exec(CREATE TRIGGER "trigger_delete_feature_count_nwm_subset_streams_levelPaths_extended" AFTER DELETE ON "nwm_subset_streams_levelPaths_extended" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count - 1 WHERE lower(table_name) = lower('nwm_subset_streams_levelPaths_extended'); END;) failed: database is locked
Querying NWM catchments ...
ERROR 1: sqlite3_exec(CREATE TABLE gpkg_extensions (table_name TEXT,column_name TEXT,extension_name TEXT NOT NULL,definition TEXT NOT NULL,scope TEXT NOT NULL,CONSTRAINT ge_tce UNIQUE (table_name, column_name, extension_name))) failed: disk I/O error
```

`delineate_hydros_and_produce_HAND.sh`
```
Exception ignored in: <built-in function delete_DataSource>

Traceback (most recent call last):

  File "/usr/lib/python3/dist-packages/osgeo_utils/gdal_polygonize.py", line 213, in gdal_polygonize

    dst_ds = None

RuntimeError: sqlite3_exec(CREATE TRIGGER "trigger_delete_feature_count_catchments" AFTER DELETE ON "catchments" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count - 1 WHERE lower(table_name) = lower('catchments'); END;) failed: database is locked

May be caused by: sqlite3_exec(CREATE TRIGGER "trigger_insert_feature_count_catchments" AFTER INSERT ON "catchments" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count + 1 WHERE lower(table_name) = lower('catchments'); END;) failed: data
```

### LINKED ISSUE #1857: [8pt] Get DEMS failed on polygonize
During an attempt to run acquire_and_preprocess_3dep_dems.py, it failed when attempting to polygonize. While I did not review the results, it did complete the stage of pulling down the DEMS then while at HUC 03070107.

I was not watching the performance monitors when it happened.

Image

Image

mluck:
Fixes SQLite errors caused by geopackage file handling by the OS (introduced in #1805) by using GeoParquet files (sorted by Hilbert curve) created by Python instead of system GDAL commands. Writing to GeoParquet is simplified by adding the capability to `to_file()`. Three new files replace the three GDAL command-line programs `gdal_polygonize.py`, `gdal_rasterize.py`, and `ogr2ogr`. In a couple of intermediate cases where GDAL was called directly and does not accept geoparquet input files, shapefiles were used instead of geopackages.

Bash scripts were also refactored to use bash arrays and use of explicit notation including curly braces around environment variables, quotes around arguments, and double brackets around conditional blocks.

Closes #1853 and closes #1857.

### Additions
- `sitecustomize.py`: Redefines geopandas.to_parquet() to automatically sort by Hilbert curve, then adds that functionality to geopandas.to_file() so that GeoParquet files can be written with `to_file()` ignoring certain keyword argumens such as `layer`, and `overwrite` and sets `driver='fiona'` as the default for Geopackage (.gpkg) files. Also specifies the use of `spawn` to force clean process spawning for multiprocessing and prevent segmentation faults.
- `src/`
    - `subset_vectors_to_branches.py`: Replaces GDAL command-line `ogr2ogr`
    - `utils/`
        - `polygonize_raster.py`: Converts from raster to vector as a replacement for the GDAL command-line `gdal_polygonize.py`
        - `rasterize_parquet.py`: Converts from vector to raster as a replacement for the GDAL command-line `gdal_rasterize.py`

##### The files below previously wrote GeoParquet files but were modified to use the new `to_file()`.
- `data/`
    - `buildings/get_fema_buildings.py` and `make_buildings_parts_per_huc.py`
    - `nws/ahps_bench_polys_to_calb_pts.py` and `merge_nws_usgs_point_parquet.py`
    - `slope/sword_slope_create_parquet_qc.py`
    - `usgs/acquire_and_preprocess_3dep_dems.py` and `write_parquet_from_calib_pts.py`
- `src/src_adjust_spatial_obs.py`
- `tools/catfim/generate_categorical_fim.py`

#### Changed to use geoparquet instead of geopackages and refactored
- `src/`
    - `delineate_hydros_and_produce_HAND.sh`, `reachID_grid_to_vector_points.py`, `run_by_branch.sh`, `run_huc.sh`

#### Changed to use geoparquet instead of geopackages
- `config/`
    - `deny_branch_zero.lst`, `deny_branches.lst`, `deny_unit.lst`
- `src/`
    - `add_crosswalk.py`, `adjust_floodplains.py`, `aggregate_branches_to_huc.py`, `associate_levelpaths_with_levees.py`, `derive_level_paths.py`, `filter_catchments_and_add_attributes.py`, `heal_bridges_osm.py`, `longitudinal_flow_adjustment.py`, `make_stages_and_catchlist.py`, `mask_dem.py`, `mitigate_branch_outlet_backpool.py`, `process_buildings_fimpact.py`, `process_roads_fimpact.py`, `split_flows.py`, `src_adjust_ras2fim_rating.py`, `src_adjust_spatial_obs.py`, `src_adjust_usgs_rating_trace.py`, `src_roughness_optimization.py`, `stream_branches.py`, `usgs_gage_crosswalk.py`, `usgs_gage_unit_setup.py`
 - `tools/
     - catfim/`
         - `vis_categorical_fim.py`
     - `bridge_inundation.py`, `evaluate_crosswalk.py`, `rating_curve_comparison.py`

#### Refactored
- `fim_pre_processing.sh`, `fim_pipeline.sh`, `fim_process_huc.sh`, `fim_postprocessing.sh`
- `src/`
    - `bash_functions.env`, `calibrate_rating_curves.sh`, `process_branch.sh`

### Testing
Tested on HUCs 02050101, 03100204, 07120006, 12040101, and 12090301.

    - [ ] Yes
    - [x] No  (f no.. skip the rest of the Deployment Plan section)
    
       - [ ] Yes
       - [ ] No
       - [ ] No
       - [ ] Yes
           -  Require new pre-clip set or any other data reloads, such as DEMS, osm, etc. ie.. pre-requisite re-data upstream of your input changes.
                - [ ] Yes
                - [ ] No
           -  Has the inputs been copied/exist in all four enviros:
                 - [ ] FIM EFS  
                 - [ ] FIM S3
                 - [ ] ESIP
                 - [ ] Dev1
            
    - [ ] Yes

- [x] Where applicable, has fim_pipeline been tested with multiple HUCs, including some other unaffected HUCs?

### Reviewer / Approver Checklist
- [ ] Where applicable, has fim_pipeline been tested with muliple HUCs, including some other unaffected HUCs?
- [ ] If there are new inputs, have you confirmed that they have been copied to all enviroments?


mluck:
### Testing results
- Tested the bash array syntax against the logging system by giving it an invalid argument flag. Also tested against an incorrect subprocess command (error log copied below). These new errors were caught and logged identical to the same error introduced in v4.9.20.1.
{6D482F27-5A49-43A3-88AD-521DCABA7DB7}


- Ran a UAT pipeline and alpha test. No errors in logs other than Exit Status 61, 63, and 64. Metrics were identical to about the 5th decimal place and differences could be due to the precision of geopackages vs. geoparquet and the effect on geometries.
{6DAF45D3-B300-4163-9CD7-8512D6792BF6}
{3A965F1D-C025-4400-83C9-855757B1EA48}

AliForghani-NOAA:
In `src/delineate_hydros_and_produce_HAND.sh:216`, the "RASTERIZE LANDSEA (OCEAN AREA) POLYGON" step still calls `gdal_rasterize` directly on `LandSea_subset.gpkg`, unlike every other rasterize call in this script which was converted to use `rasterize_parquet.py`.

Was this intentionally left as gdal_rasterize/GPKG, or should it also be converted to .parquet + rasterize_parquet.py for consistency with the rest of this PR?

mluck:
> In `src/delineate_hydros_and_produce_HAND.sh:216`, the "RASTERIZE LANDSEA (OCEAN AREA) POLYGON" step still calls `gdal_rasterize` directly on `LandSea_subset.gpkg`, unlike every other rasterize call in this script which was converted to use `rasterize_parquet.py`.
> 
> Was this intentionally left as gdal_rasterize/GPKG, or should it also be converted to .parquet + rasterize_parquet.py for consistency with the rest of this PR?

In this case `gdal_rasterize` is appropriate because the data file is coming directly from `/inputs` as a GPKG whereas `rasterize_parquet.py` is intended for rasterizing GeoParquet files that are created during FIM pipeline.

AliForghani-NOAA:
I see `tools/rating_curve_comparison.py` [lines 828-841](https://github.com/NOAA-OWP/inundation-mapping/blob/93b53274e631205d21a5db1af672538aae1eb36a/tools/rating_curve_comparison.py#L832-L839) still reads `demDerived_reaches_split_filtered_addedAttributes_crosswalked_{branch}.gpkg`. Since `add_crosswalk.py` now writes this file as .parquet only, need to update this consumer to gpd.read_parquet(...) on the .parquet path instead?

src/add_crosswalk.py
```diff
@@ -459,10 +463,19 @@ def add_crosswalk(
         output_src_json[str(hid)] = {'q_list': q_list, 'stage_list': stage_list}
 
     # write out
+    to_hilbert_parquet(output_catchments, output_catchments_fileName, index=False)
+
+    # HACK -- save a duplicate in .gpkg for now
```
AliForghani-NOAA: Please add a more descriptive comment explaining why this output also needs to be saved as a GeoPackage file (e.g., several downstream tools, such as `tools/compute_flood_depth.py`, still expect this data in GeoPackage format).

AliForghani-NOAA:
It looks like `src/utils/polygonize_raster.py` no longer has any callers after the addition of `src/polygonize_raster.py`. Its only caller, `data/usgs/acquire_and_preprocess_3dep_dems.py`, now uses the new file. Can we remove `src/utils/polygonize_raster.py`?

src/heal_bridges_osm.py
```diff
@@ -180,7 +182,7 @@ def process_bridges_in_huc(
     # Check if the GeoDataFrame is empty
     if not osm_gdf.empty:
         # Write the bridge points to a geopackage
-        osm_gdf.to_file(bridge_centroids, index=False, engine='fiona')
+        to_hilbert_parquet(osm_gdf, bridge_centroids, index=False)
```
AliForghani-NOAA: Since the `osm_bridge_centroid` outputs are now saved in Parquet format, the [aggregation logic](https://github.com/NOAA-OWP/inundation-mapping/blob/93b53274e631205d21a5db1af672538aae1eb36a/src/aggregate_branches_to_huc.py#L259) should be updated accordingly. Similarly, this [tool ](https://github.com/NOAA-OWP/inundation-mapping/blob/8388722ae96848ace724e67ea5a70bbe1e1ca280/tools/bridge_inundation.py#L72-L78)needs to be updated. 

@RobHanna-NOAA Can you confirm HV is ok with using Parquet format for `osm_bridge_centroids` outputs?

RobHanna-NOAA: Yes.. parquet preferred in all scenerios

src/add_crosswalk.py
RobHanna-NOAA: added note to code inline.

mluck:
> It looks like `src/utils/polygonize_raster.py` no longer has any callers after the addition of `src/polygonize_raster.py`. Its only caller, `data/usgs/acquire_and_preprocess_3dep_dems.py`, now uses the new file. Can we remove `src/utils/polygonize_raster.py`?

Yes, those were nearly duplicate files. I deleted `src/polygonize_raster.py` in favor of keeping `src/utils/polygonize_raster.py` and updated the relevant files.

src/heal_bridges_osm.py
mluck: `aggregate_branches_to_huc.py` and `tools/bridge_inundation.py` have been updated.

src/add_crosswalk.py
mluck: @RobHanna-NOAA added a descriptive note

mluck:
> I see `tools/rating_curve_comparison.py` [lines 828-841](https://github.com/NOAA-OWP/inundation-mapping/blob/93b53274e631205d21a5db1af672538aae1eb36a/tools/rating_curve_comparison.py#L832-L839) still reads `demDerived_reaches_split_filtered_addedAttributes_crosswalked_{branch}.gpkg`. Since `add_crosswalk.py` now writes this file as .parquet only, need to update this consumer to gpd.read_parquet(...) on the .parquet path instead?

Done.

tools/rating_curve_comparison.py
```diff
@@ -829,14 +829,14 @@ def generate_rc_and_rem_plots(rc, plot_filename, recurr_data_table, branches_fol
             os.path.join(
                 branches_folder,
                 branch,
-                f'demDerived_reaches_split_filtered_addedAttributes_crosswalked_{branch}.gpkg',
+                f'demDerived_reaches_split_filtered_addedAttributes_crosswalked_{branch}.parquet',
             )
         ):
             reaches = gpd.read_file(
```
AliForghani-NOAA: This is still using `gpd.read_file`, even though the input file has been changed to Parquet.

tools/catfim/vis_categorical_fim.py
```diff
@@ -73,7 +73,7 @@ def read_catfim_outputs(catfim_inputs_path, catfim_outputs_path, huc):
     '''
 
     # Read in HAND output flowlines
-    flowline_path = os.path.join(catfim_inputs_path, huc, 'nwm_subset_streams_levelPaths_dissolved.gpkg')
+    flowline_path = os.path.join(catfim_inputs_path, huc, 'nwm_subset_streams_levelPaths_dissolved.parquet')
     flowline_gdf = gpd.read_file(flowline_path)
```
AliForghani-NOAA: This is still using `gpd.read_file`, even though the input file has been changed to Parquet.

AliForghani-NOAA:
I still see two instances in `src/aggregate_branches_to_huc.py` where the bridge file format was changed to Parquet, but the Parquet I/O API is not being used: [here](https://github.com/NOAA-OWP/inundation-mapping/blob/434f493bc71b04750c518789ade0ab03fd9fb281/src/aggregate_branches_to_huc.py#L475) and [here](https://github.com/NOAA-OWP/inundation-mapping/blob/434f493bc71b04750c518789ade0ab03fd9fb281/src/aggregate_branches_to_huc.py#L263).

AliForghani-NOAA:
I also found the following issues:

1. https://github.com/NOAA-OWP/inundation-mapping/blob/8b7acb6e2cc470cc764a4b34e9aa18dea4453910/src/mitigate_branch_outlet_backpool.py#L522-L523

   * output_flows.to_file(split_flows_filename, driver="GPKG", index=False)
     `split_flows_filename` has a `.parquet` extension, so this should use the GeoParquet I/O API.

   * split_points_filtered_geom.to_file(split_points_filename, driver="GPKG", index=False)
     `split_points_filename` has a `.shp` extension, so the driver should be `"ESRI Shapefile"` or omitted altogether.

2. There are several related issues involving `catchments_poly_path` in `src_roughness_optimization.py`:

   **2.1. Caller inconsistency**
   `src/src_adjust_usgs_rating_trace.py` calls `update_rating_curve()` with `catchments_poly_path` pointing to a `.parquet` file, while the other two callers, `src_adjust_ras2fim_rating.py` and `src_adjust_spatial_obs.py`, still pass `.gpkg` files.

   This inconsistency is not necessarily a bug by itself, but it exposes the format-handling issues below.

   **2.2. Incorrect write API in `update_rating_curve()`**
   In `src/src_roughness_optimization.py`, `update_rating_curve()` correctly handles both GeoPackage and GeoParquet when reading `catchments_poly_path` by checking the file extension. However, it later writes the data unconditionally using (around lines 471–478 and 493–500):

   ```python
   to_file(catchments_poly_path, driver="GPKG", ...)
   ```

   This would cause a bug in `src_adjust_usgs_rating_trace.py` because it passes a `.parquet` path.


   **2.3. Ungated read operation**
   There is also an unconditional `gpd.read_file(catchments_poly_path)` at [src/src_roughness_optimization.py:527](https://github.com/NOAA-OWP/inundation-mapping/blob/8b7acb6e2cc470cc764a4b34e9aa18dea4453910/src/src_roughness_optimization.py#L527), that would break for the .parquet caller--This should also be gated by the file extension.

3. In `tools/evaluate_crosswalk.py`, compare [line 179](https://github.com/NOAA-OWP/inundation-mapping/blob/8b7acb6e2cc470cc764a4b34e9aa18dea4453910/tools/evaluate_crosswalk.py#L179) with [line 99](https://github.com/NOAA-OWP/inundation-mapping/blob/8b7acb6e2cc470cc764a4b34e9aa18dea4453910/tools/evaluate_crosswalk.py#L99). 
   `input_flows_fileName` points to a Parquet file, so line 179 should be updated to use the same Parquet-reading approach used at line 99.

mluck:
I tested this PR against the original issue #1853 by using EFS as the temp write folder during a FIM pipeline run on two small HUCs (03100204 and 05030104). `dev-geoparquet` finished with zero errors while v4.9.21.2 had two SQL errors. In addition, the `dev-geoparquet` outputs finished in 14 minutes 30 seconds with a total folder size of 383M while v4.9.21.2 took 17:30 and resulted in 711M folder size.

AliForghani-NOAA:
Well done @mluck on the many improvements. I have one final suggestion regarding the outputs of all, or most, scripts under `tools/` that are not part of the FIM pipeline.

It would be helpful to allow users to choose either GPKG or Parquet as the output format. The implementation below is a perfect example: https://github.com/NOAA-OWP/inundation-mapping/blob/d18be16ebedd5814ae46af6f0900995cb61b218b/tools/fimpacts_inundation.py#L191-L194

Then, the `-o` help text [here](https://github.com/NOAA-OWP/inundation-mapping/blob/d18be16ebedd5814ae46af6f0900995cb61b218b/tools/fimpacts_inundation.py#L219) should also be updated to read:

```text
Path to GPKG or Parquet output.
```

The same consideration applies to `tools/bridge_inundation.py`. The output logic [here](https://github.com/NOAA-OWP/inundation-mapping/blob/d18be16ebedd5814ae46af6f0900995cb61b218b/tools/bridge_inundation.py#L121) should support writing either GPKG or Parquet based on the output file extension, and the `-o` help text should be updated accordingly.

AliForghani-NOAA:
> I tested this PR against the original issue #1853 by using EFS as the temp write folder during a FIM pipeline run on two small HUCs (03100204 and 05030104). `dev-geoparquet` finished with zero errors while v4.9.21.2 had two SQL errors. In addition, the `dev-geoparquet` outputs finished in 14 minutes 30 seconds with a total folder size of 383M while v4.9.21.2 took 17:30 and resulted in 711M folder size.

This looks great—nice to see that the PR not only resolves the SQL errors but also improves runtime and significantly reduces the output size. Great work!

src/utils/shared_functions.py
```diff
@@ -32,6 +34,59 @@
...
+    x_span = (maxx - minx) if maxx != minx else 1.0
+    y_span = (maxy - miny) if maxy != miny else 1.0
+    max_val = (1 << p) - 1
+
+    # 2. Rescale coordinates to integer arrays (using .to_numpy() to avoid index misalignment)
+    norm_x = ((rep_points.x.to_numpy() - minx) / x_span * max_val).clip(0, max_val).astype(int)
+    norm_y = ((rep_points.y.to_numpy() - miny) / y_span * max_val).clip(0, max_val).astype(int)
+
+    # 3. Compute Hilbert distances
+    hc = HilbertCurve(p=p, n=2)
```
groutr: GeoPandas can natively sort by hilbert distance.

```
gdf.sort_values(by="geometry")
```
https://github.com/geopandas/geopandas/pull/2786

AliForghani-NOAA:
As @groutr mentioned, `sort_values(by="geometry")` applies GeoPandas’ default Hilbert ordering. However, the `p` parameter in [to_hilbert_parquet()](https://github.com/NOAA-OWP/inundation-mapping/blob/d72ac08588de23ff0271e4719765e61f1b2dd972/src/utils/shared_functions.py#L39) function is currently unused. If we want to retain this parameter, could we calculate the Hilbert distance explicitly using `p` and sort by the resulting values (like below)? This would make the intended behavior clearer and ensure that the requested Hilbert precision is applied.

```
hilbert_distance = gdf.geometry.hilbert_distance(level=p)
hilbert_order = hilbert_distance.to_numpy().argsort(kind="stable")
gdf.iloc[hilbert_order].to_parquet(output_path, **to_parquet_kwargs)
```

groutr:
@AliForghani-NOAA Is there a need to keep the `p` parameter? I would suggest its removal to simplify the function signature.

AliForghani-NOAA:
> @AliForghani-NOAA Is there a need to keep the `p` parameter? I would suggest its removal to simplify the function signature.

Agreed. Alternatively, we can remove `p` to simplify the function, since we do not currently have a use case for custom Hilbert precision.

src/utils/io.py
```diff
@@ -0,0 +1,41 @@
...
+    # Route 1: Parquet -> Hilbert sort + to_parquet()
+    if not has_layer and (driver == "parquet" or (driver is None and ext == ".parquet")):
+        parquet_kwargs = {
+            k: v
+            for k, v in kwargs.items()
+            if k not in ("driver", "layer", "schema", "mode", "engine", "overwrite")
+        }
+
+        if not gdf.empty and gdf.active_geometry_name is not None and not gdf.geometry.is_empty.all():
+            gdf = gdf.sort_values(by="geometry", ascending=True).reset_index(drop=True)
```
groutr: What is the purpose of dropping the index here?

src/utils/io.py
```diff
@@ -0,0 +1,41 @@
...
+    has_layer = "layer" in kwargs and kwargs["layer"] is not None
+
+    driver_arg = kwargs.get("driver")
+    driver = driver_arg.lower() if isinstance(driver_arg, str) else None
+
+    ext = filepath.suffix.lower() if filepath else ""
+
+    # Route 1: Parquet -> Hilbert sort + to_parquet()
+    if not has_layer and (driver == "parquet" or (driver is None and ext == ".parquet")):
+        parquet_kwargs = {
```
groutr: This could more succinctly be written as
```python
exclude = ("driver", "layer", "schema", "mode", "engine", "overwrite")
parquet_kwargs = {k: kwargs[k] for k in kwargs.keys() - exclude}
```

mluck:
It discards the non-sequential row numbers created by Hilbert sorting. In most cases we don’t use a named index and the row identifier is in a column like “HydroID”. I added a check for a named index that will be kept if one exists.

src/utils/io.py
```diff
@@ -0,0 +1,53 @@
...
+        parquet_kwargs.setdefault("write_covering_bbox", True)
+        parquet_kwargs.setdefault("compression", "zstd")
+        parquet_kwargs.setdefault("write_page_index", True)
+        parquet_kwargs.setdefault("write_page_checksum", True)
+
+        if not gdf.empty and gdf.active_geometry_name is not None and not gdf.geometry.is_empty.all():
+            # Sort spatially on Hilbert curve
+            gdf = gdf.sort_values(by="geometry", ascending=True)
+
+            if gdf.index.name is None:
```
groutr: With ignore_index, this check can become:
```python
if ignore_index:
    gdf = gdf.reset_index(drop=True)
```

src/utils/io.py
```diff
@@ -0,0 +1,53 @@
...
+                gdf = gdf.reset_index(drop=True)
+
+        # Any write error is surfaced to the caller.
+        gdf.to_parquet(filename, **parquet_kwargs)
+
+    # Route 2: GeoPackage -> to_file()
+    elif driver in ("gpkg", "geopackage") or (driver is None and ext == ".gpkg"):
```
groutr: Reading the documentation of to_file, it doesn't seem necessary to special case gpkg. This seems to be duplicating how to_file already behaves. I think it would be sufficient to just have:
```python
else:
    gdf.to_file(filename, *args, **kwargs)
```

src/utils/io.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # ROUTE 2: All Vector File Formats (GPKG, Shapefile, GeoJSON, GDB, etc.)
+    else:
+        out_kwargs = kwargs.copy()
+
+        # If writing a GeoPackage, apply Fiona defaults
+        if driver in ("gpkg", "geopackage") or (driver is None and ext == ".gpkg"):
+            out_kwargs.setdefault("engine", "fiona")
+
+        if ignore_index:
```
AliForghani-NOAA: Since geopandas' `to_parquet()/to_file()` already accept an `index` arg (for dropping the index altogether before writing), adding a new param with a similar name like `ignore_index` on `write_geodataframe()` is confusing and unnecessary. I suggest removing the `ignore_index` parameter entirely, and instead checking the caller-provided `index` arg — if it's not False (caller wants to keep the index), automatically reset it, like below:
```python
if kwargs.get('index') is not False:
    gdf = gdf.reset_index(drop=True)
```

AliForghani-NOAA:
Noticed the `spawn` fix (`mp.set_start_method("spawn", force=True)`) disappeared when `sitecustomize.py` was deleted in favor of `io.py`. Was that dropped intentionally? If it's still wanted a small, single-purpose `sitecustomize.py` containing only the `spawn` call (nothing else) seems like the right scope for it, rather than adding it to every multiprocessing call site individually.

RobHanna-NOAA:
Yes.. in my inudation fix PR, there are a bunch of discoveries around processpools. I will get a list of them all together and we can fix them in future PRs.
But, one that I found that turns out to be critical for memory management is the spawn you mentioned above. If we are playing with spawning on MP, might as well as add that one for now too.

ie)
image

mluck:
> Noticed the `spawn` fix (`mp.set_start_method("spawn", force=True)`) disappeared when `sitecustomize.py` was deleted in favor of `io.py`. Was that dropped intentionally? If it's still wanted a small, single-purpose `sitecustomize.py` containing only the `spawn` call (nothing else) seems like the right scope for it, rather than adding it to every multiprocessing call site individually.

I elevated `spawn` to its own issue (#1920).

AliForghani-NOAA:
Great work on this PR. I’m all set with my review — thanks for the thoughtful work here.

AliForghani-NOAA (dismissed):

CarsonPruitt-NOAA (approved):

PR #1934, opened 2026-08-25
PR: Replace numba with numpy in lofi.

groutr:
Replaces a few of the numba routines with vectorized numpy versions.

Benchmarks for the inundation code shows the numpy version to be about 40% faster.

For HUC 18100100
numba:
17.5 s ± 23.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

numpy:
12.7 s ± 26.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

This should effectively remove the runtime requirement for numba when using LoFi. There are two other files that use numba, but refactoring them to use numpy is not as straight forward. My speculation is that numba is not really a benefit in those two other scripts.
The two other files are make_rem.py and adjust_thalweg_lateral.py

### DIFF
--- a/tools/inundation.py
+++ b/tools/inundation.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
 
 import argparse
-import os
 from os.path import splitext
-from typing import List, Optional, Tuple, Union
+from typing import Optional, Union
 from warnings import warn
 
 import fiona
@@ -12,7 +11,6 @@ import numpy as np
 import pandas as pd
 import rasterio
 import xarray as xr
-from numba import njit, typed, types
 from rasterio.mask import mask
 from shapely.geometry import shape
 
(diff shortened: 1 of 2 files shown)


RobHanna-NOAA:
@groutr : Looks great.. love all of the input.  @GregoryPetrochenkov-NOAA :  Can you help Ryan again with buliding out the PR for our patterns? (PR text, changelog, linting, latest dev, etc)?

GregoryPetrochenkov-NOAA:
So most of everything you need is in our [CONTRIBUTING.md](https://github.com/NOAA-OWP/inundation-mapping/blob/dev/CONTRIBUTING.md).  @RobHanna-NOAA This should be updated to have the PR Title you prefer, CHANGELOG.md text and etc.  

The preference for PR titles should be like the following:
[1pt] PR: Adjust non-monotonic check in LoFI subdivision

Make sure it has the latest dev branch merged in.

Install pre-commit and that will take care of the linting.  After you commit  sometimes the pre-commit scripts will not run, so either a manual linting with it's tools and/or resetting and re-committing will help.  You will see that the linting checks pass in the PR.

Update the CHANGELOG.MD as follows:

## v4.9.x.x - 2026-08-21 -[PR#1930](https://github.com/NOAA-OWP/inundation-mapping/pull/1930)

This PR stops a non-monotonic check from occurring with branch zero, adjusts the attributes, and allows all time slices of gcs ensembles to be downloaded.

### Changes
- `tools/lofi/probabilistic_get_ensembles_gcs.py`: Allow for all time slices to be downloaded.
- `tools/lofi/probabilistic_get_inundation.py`: Stop non-monotonic check on branch zero and adjust the attributes used.
- `tools/lofi/probabilistic_verison.py`: Update version.
<br />

Let me know if you have any questions regarding this process.

Issue #1879, opened 2026-07-06
[5pt] Connection error in FEMA NFHL download


mluck:
An error occurred when running `data/nfhl/download_fema_nfhl.py` for HUC 02030105:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connectionpool.py", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connectionpool.py", line 488, in _make_request
    raise new_e
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connectionpool.py", line 464, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connectionpool.py", line 1093, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connection.py", line 796, in connect
    sock_and_verified = _ssl_wrap_socket_and_match_hostname(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/urllib3/connection.py", line 975, in _ssl_wrap_socket_and_match_hostname
    ssl_sock = ssl_wrap_socket(
               ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/urllib3/util/ssl_.py", line 483, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/urllib3/util/ssl_.py", line 527, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ssl.py", line 455, in wrap_socket
    return self.sslsocket_class._create(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1937, opened 2026-08-27
fix: Add retry logic for transient connection errors in S3 bucket check (closes #1879)

botbikamordehai2-sketch:
## What

Fixes #1879. When running `data/nfhl/download_fema_nfhl.py`, a `ConnectionResetError` from the SSL layer can occur due to transient network issues. The current code does not handle such transient connection errors, causing the script to fail immediately.

## Fix

Added retry logic with exponential backoff in the `does_s3_bucket_exist` function to catch `ConnectionError` (which includes `ConnectionResetError`) and retry the `head_bucket` call up to 5 times before giving up. Also corrected `strip("")` to `strip('/')` on the bucket name to properly strip slashes.

The retry logic helps handle transient connection resets from S3, making the download process more robust.

Closes #1879


RobHanna-NOAA:
Closed. Generated by AI

